PR #106, opened 2014-12-16
Vimshell strategy

p0deje:
Strategy on top of [vimshell](https://github.com/Shougo/vimshell.vim).

Brings the following on the table (especially cool for GUI):

1. ANSI color support.
2. Allows using interactive commands (e.g. irb) and sending SIGINT (with `<C-c>`).
3. Shows command output while it's running.
4. A lot more that vimshell allows to do.

The following also concern me:

1. ~~No pid (is this a problem?).~~ Pid is now buffer number.
2. `dispatch#copen(0)` is called directly in `postexit` hook. I suppose it is not the best way to do that.
3. No support for background builds because hooks are not trigger for hidden vimshell buffer.
4. ~~No `dispatch#vimshell#activate()` at the moment but I think it can be easily added, I just need to understand the expected behaviour.~~ Activate works correctly.
5. Changes `updatetime` to low values while process is running. Vimshell adds autocmd on `CursorHold,CursorHoldI` to update output when vimshell buffer is not focused. Low `updatetime` works fine for me in GUI, but in Terminal this makes cursor blink like a crazy one. Any ideas on how to improve that?
6. Command exit code is not correct (it uses exit code from `tee`). I believe it is not really important and not all strategies implement that correctly but it can be fixed with a bit more bash.

I have implemented my own function to strip color codes from piped file with perl, but I know tmux strategy does the same, so it might make sense to share to code.

### DIFF
--- a/autoload/dispatch/vimshell.vim
+++ b/autoload/dispatch/vimshell.vim
@@ -1,0 +1,92 @@
+" dispatch.vim vimshell strategy
+
+if exists('g:autoloaded_dispatch_vimshell')
+  finish
+endif
+let g:autoloaded_dispatch_vimshell = 1
+
+function! dispatch#vimshell#handle(request) abort
+  if !exists('g:loaded_vimshell') || a:request.background
+    return 0
+  endif
+
+  let s:tempfile = a:request.file
+  let s:action = a:request.action
+
+  let vimshell_context = {
+        \ 'buffer_name': 'vim-dispatch',
+        \ 'create': 1,
+        \ 'prompt': '$ '
+        \ }
+
+  if s:action ==# 'make'
+    let vimshell_context.split = 1
+    let vimshell_context.split_command = 'botright split | resize 10'
+  elseif s:action ==# 'start'
+    let vimshell_context.tab = 1
+    execute 'tabnew'
+  endif
+
+  call vimshell#init#_start(a:request.directory, vimshell_context)
+  call vimshell#hook#add('preparse', 'dispatch', 'dispatch#vimshell#preparse')
+
+  if s:action ==# 'make'
+    call vimshell#hook#add('postexit', 'dispatch', 'dispatch#vimshell#postexit')
+    execute 'wincmd p'
+  endif
+
+  let a:request.pid = bufnr('%')
+  let s:pid = a:request.pid
+
+  call vimshell#interactive#send(a:request.expanded)
+
+  return 1
+endfunction
+
+function! dispatch#vimshell#activate(pid) abort
+  for tabpage in range(tabpagenr('$'))
+    let tabnumber = tabpage + 1
+    for bufnr in tabpagebuflist(tabnumber)
+      if bufnr ==# a:pid
+        execute 'tabnext '.tabnumber
+        execute bufwinnr(bufnr).'wincmd w'
+
+        return 1
+      endif
+    endfor
+  endfor
+endfunction
+
+function! dispatch#vimshell#running(pid) abort
+  return bufexists(str2nr(a:pid))
+endfunction
+
+function! dispatch#vimshell#preparse(...) abort
+  call vimshell#hook#remove('preparse', 'dispatch')
+
+  setlocal norelativenumber
+  setlocal nonumber
+
+  let command = a:1
+
+  if s:action ==# 'make'
+    let s:updatetime = &updatetime
+    let &updatetime = 100
+    let command .= ' '.&shellpipe.' '.s:tempfile
+  endif
+
+  return 'clear; '.
+        \ 'echo '.s:pid.' > '.s:tempfile.'.pid; '.
+        \ &shell.' '.&shellcmdflag.' '.vimproc#shellescape(command).
+        \ '; exit'
+endfunction
+
+function! dispatch#vimshell#postexit(...) abort
+  call vimshell#hook#remove('postexit', 'dispatch')
+  let &updatetime = s:updatetime
+
+  call system(&shell.' '.&shellcmdflag.' '.
+        \ shellescape("perl -pi.bak -e 's/\x1b.*?[mGKH]//g' ".s:tempfile))
+  call dispatch#copen(0)
+  execute 'wincmd p'
+endfunction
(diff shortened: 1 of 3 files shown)


p0deje:
@tpope Is there anything else I should do to get it merged?

rgrinberg:
@tpope plx merge

rgrinberg:
I use VimShell and would love to see this.

tpope:
Still needs activate support, no? This is what causes `:Start` to focus the window when you run it with the same command a second time. Guessing in this case you want to focus the buffer.

p0deje:
> Regarding the activate - what is expected behavior for this? I cannnot figure out since :Start opens new full screen split (so there is no way to activate it since Vim is in background) and :Dispatch always open new process/split for the same command.

I asked this because that's how Tmux strategy works now, so I duplicated the behavior. I'll see if I can make `:Start` work as you propose.

p0deje:
I am not sure how to implement this. I suppose it should work the following way:

1. `:Start irb` opens new tab (like pane in Tmux) and spawns process.
2. Switch back other tab, do some code.
3. Execute `:Start irb` and it should switch to correct tab and window of running process.

The problem is that if I save references to tabpage number, window number and buffer number, it's still not enough. For instance, tabpage number may be changed if new tab is opened, so it's not reliable to just do `:tabnext {tabpagenumber}`.

The other approach would be not to create new tab, just replace current buffer with `:Start` and then simply do `:b {buffernumber}`. However, I am not sure it's very convenient.

Do you have any suggestions on what's the best way to implement this?

tpope:
You could iterate through `tabpagebuflist()` for each tab to hunt down the buffer.

p0deje:
Thanks for `tabpagebuflist()` - I didn't know about that.

`:Start` with activation support now works correctly, but there is a problem when command exits, tabpage with vimshell buffer switches to alternate buffer, not close (Shougo/vimshell.vim#192). Once this is fixed or worked around, I think this pull request finished (at least, from my point of view).

It would be great if you could check it out (I already mentioned the problem with `updatetime` in terminal Vim), leave your comments and review the code.

tpope:
I get errors when I try dispatching.

```
Error detected while processing function <SNR>242_check_all_output..<SNR>242_check_output:                                                                                                     
line   74:                                                                                                                                                                                     
E121: Undefined variable: b:interactive                                                                                                                                                        
Press ENTER or type command to continue                                                                                                                                                        
Error detected while processing function <SNR>242_check_all_output..<SNR>242_check_output:                                                                                                     
line   74:                                                                                                                                                                                     
E15: Invalid expression: !is_last_line && pos != getpos('.') && b:interactive.process.is_valid                                                                                                 
```

Also, no pid (not even the pseudo-pid buffer number discussed above).

p0deje:
Right, this happens on the latest version of Vimshell, though I was testing few commits before.

I'll look into this.

p0deje:
> I get errors when I try dispatching.

Depends on Shougo/vimshell.vim#193. I've posted workaround, looking forward to see if it is merged.

> Also, no pid (not even the pseudo-pid buffer number discussed above).

It was saved internally for `:Start` logic, but not echoed. Fixed now.

petobens:
+1

p0deje:
@tpope All the problems are now resolved with latest vimshell.vim, can you take a look one more time?

tpope:
I'm not a fan of how it shows the command with a shell prompt before the output. It makes it look like one of those tmux send-keys runners. Is that fixable?

I'm also feeling iffy about hacks like mucking around with `'updatetime'`, and am wondering if this would be better suited as a separate extension. I would definitely make the requisite internal changes if going that route.

p0deje:
> I'm not a fan of how it shows the command with a shell prompt before the output. It makes it look like one of those tmux send-keys runners. Is that fixable?

That's definitely fixable, I can do that if you want though I kinda like the fact it shows what it is running.

> I'm also feeling iffy about hacks like mucking around with 'updatetime', and am wondering if this would be better suited as a separate extension. I would definitely make the requisite internal changes if going that route.

I believe it makes sense especially since vimshell is hacky enough to introduce new bugs sometimes and it might be painful to support it within vim-dispatch (especially since you're not using vimshell, so I'll have to submit pull requests all the time). I would do that as a separate plugin, but the internal changes are very specific to vimshell (e.g. `dispatch#vimshell#pid_format()`). I believe there has to be done some refactoring to dispatch to make it fully extensible.

tpope:
I think the place to show the command being run is the statusline.

    let &l:statusline = substitute(request.expanded, '%', '%%', 'g')

The `handler-running` branch has an attempt to make `s:running()` overridable by handlers. Just define `#running` in your handler. If it's sufficient I'll merge it in.

p0deje:
> I think the place to show the command being run is the statusline.

Unfortunately, vimshell overrides statusline on buffer change. I'll probably raise an issue for that, but so far I've just implemented it so it doesn't show what it's running.

> The handler-running branch has an attempt to make s:running() overridable by handlers. Just define #running in your handler. If it's sufficient I'll merge it in.

Okay, I can implement `dispatch#vimshell#running()`, but what about the changes in `dispatch#pid()`?

tpope:
Is dropping the ` && getfsize(file.'.pid') > 0` clause near the end enough to make that work?

tpope:
For that, matter, could you just dump the buffer number in that .pid file? Or does vimshell give you any way to get at the pid?

p0deje:
Okay, so now it just dumps buffer number of `.pid` file. I also extracted `#running()` function and changed the behavior so now it doesn't show prompt and running command.

tpope:
I guess my vote is to make a separate plugin for now. It shouldn't take it much more than a `autocmd VimEnter *` event to `insert()` at the appropriate `index()` in `g:dispatch_handlers`.

I may reassess after the next dispatch.vim release. Link me when your plugin is up on GitHub. I'll install it so I can spend some real time using it.

p0deje:
Okay, I've release [vim-dispatch-vimshell](https://github.com/p0deje/vim-dispatch-vimshell) which you can try and use.

Thanks for all the feedback and guidance. Hope it will be the part of vim-dispatch one day.

PR #109, opened 2014-12-26
Fixed #108 : E484 when used on Windows GVim with cygwin shell

### LINKED ISSUE #108: Cause E484 when used on Windows GVim with cygwin shell
When `&shell` is set to cygwin shell (in my case it's `<path to cygwin>/bin/zsh.exe`), using this plugin in windows gvim gives something like `E484: Can't open file <temp-dir>/VIo3EEB.tmp`, while the requested command is correctly executed. This `<temp-dir>` path does exists.

A temporary workaround is to change the `&shell` back into the default `cmd`, but this breaks some of my other plug-ins.

An example of error screen is as follows:
untitled
This is produced by executing `:Start python3 -m unittest -v`, while the unit tests are still correctly executed asynchronously.

cybeliak:
On Windows gvim, if the shell is set to cygwin shell or msys shell, these two `system()` call would cause the shell misinterpret the quoting of the command. Thus, I suggest to escape double quotes when these shells are used, while keeping the original command when the default `cmd` shell is used.

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -581,7 +581,11 @@ function! s:running(pid) abort
   if !a:pid
     return 0
   elseif has('win32')
-    return system('tasklist /fi "pid eq '.a:pid.'"') =~# '==='
+    let tasklist_cmd = 'tasklist /fi "pid eq '.a:pid.'"'
+    if &shellxquote ==# '"'
+      let tasklist_cmd = substitute(tasklist_cmd, '"', "'", "g")
+    endif
+    return system(tasklist_cmd) =~# '==='
   else
     call system('kill -0 '.a:pid)
     return !v:shell_error
--- a/autoload/dispatch/windows.vim
+++ b/autoload/dispatch/windows.vim
@@ -71,9 +71,14 @@ function! dispatch#windows#start(request) abort
 endfunction
 
 function! dispatch#windows#activate(pid) abort
-  if system('tasklist /fi "pid eq '.a:pid.'"') !~# '==='
+  let tasklist_cmd = 'tasklist /fi "pid eq '.a:pid.'"'
+  if &shellxquote ==# '"'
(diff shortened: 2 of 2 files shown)


tpope:
Can we use `shellescape()` instead? I've found matching against `'shell'` to be a bit brittle.

cybeliak:
`shellescape()` doesn't seem to do the trick, since it won't actually touch double quotes when used with posix shells (where `&shellescape` is usually set). To make it worse, when used with default `cmd`, this function escapes `"` into `""`, which makes sense on filenames but not on whole command which already have `"`s inside. As mentioned in `help shellescape`:

> On MS-Windows and MS-DOS, when 'shellslash' is not set, it
> will enclose {string} in double quotes and double all double
> quotes within {string}.

However, I've found that the actual problem is not in the shell itself, but `&shellxquote` which vim automatically set to a double quote when it sees a posix-like shell being used, as mentioned in `help shellxquote`:

> 'shellxquote' 'sxq'	string	(default: "";
> 			for Win32, when 'shell' is cmd.exe: "("
> 			for Win32, when 'shell' contains "sh"
> 			somewhere: "\""
> 			for Unix, when using system(): "\"")

Further, I agree that doing string matching on `&shell` smells bad. In the updated commit, the command is only altered when `&shellxquote` is a double quote, which is what actually clash with double quotes within the command, instead of depending oh matching `&shell`.

It's inspired by the `s:escape` function from `autoload/dispatch/windows.vim` (though for this task `s:escape` is doing a bit too much).

tpope:
Perfect. Tweaked the commit subject and cherry picked. Thanks!

tpope:
Recently I've made some changes to escaping and expanding to fix a bunch of edge cases. @cybeliak, if you're still an active user of dispatch.vim, verifying I didn't break anything with your setup would be great.

PR #122, opened 2015-04-22
Add foreground dispatch support for screen

vedknom:
Builds on top and following suggestions from #84 to add support for screen foreground execution. Need to add `HUP` for trap to work in non-zombie mode, and `screen -X at $WINDOW kill` is needed for zombie mode to completely remove the window.


tpope:
I guess my first question is why the `aftercmd` dance rather than just appending a string in `screen.vim`?

Issue #127, opened 2015-06-03
X11 handler is ignored in gvim


zorab47:
The X11 handler is consistently skipped when running in gvim and uses the fallback headless handler.

The check for `a:request.action` in [x11.vim:9](https://github.com/tpope/vim-dispatch/blob/2f5d15fea4c9427ba3f7d8d82c838890b8522125/autoload/dispatch/x11.vim#L9) consistently returns early when invoking dispatch. My guess is because the request's action is set to "make" it causes the handler to be skipped. Should the "make" action be handled differently in the X11 handler?

Commenting out the check for `a:request.action !=# 'start'` allows the x11 handler to be used and starts a new terminal session as expected.

tpope:
Oh, yeah, that can probably be tweaked to allow make through if `wmctrl` is available. Without `wmctrl` there is no way to return focus to Vim, which is the origin of that short circuit.

tpope:
Try the `x11-make` branch. Make sure all 4 cases (Make/Make!/Start/Start!) work as expected and I'll merge.

tpope:
Not sure what I was thinking. Someone needs to implement make support for this to work.

zorab47:
The commands Make/Make!/Start/Start! all executed in the new terminal session on the `x11-make` branch. I did notice that the call to Make didn't focus the new terminal in front of gvim like Start did. Both Make!/Start! executed in the background without showing the new terminal session.

What would be involved with implementing make support?

tpope:
Basically you need to use `prepare_make` rather than `prepare_start`. The screen adapter is probably the best source of inspiration.

zorab47:
Thanks. I'll give it a shot.

aripollak:
Looks like this can be closed now. The X11 support still seems pretty buggy to me, but I'll open followup bugs as I isolate them.

tpope:
Yeah I used it for the first time yesterday and fixed a bad bug. Let me know if you find any more.

PR #128, opened 2015-06-05
Add :Make support to X11 handler

zorab47:
Allow X11 handler to execute both :Make and :Start commands.

See #127 for initial discussion.

### DIFF
--- a/autoload/dispatch/x11.vim
+++ b/autoload/dispatch/x11.vim
@@ -6,10 +6,11 @@ endif
 let g:autoloaded_dispatch_x11 = 1
 
 function! dispatch#x11#handle(request) abort
-  if $DISPLAY !~# '^:' || a:request.action !=# 'start'
+  if $DISPLAY !~# '^:'
     return 0
   endif
-  if a:request.background && (!v:windowid || !executable('wmctrl'))
+  if (a:request.background || a:request.action ==# 'make') &&
+        \ (!v:windowid || !executable('wmctrl'))
     return 0
   endif
   if !empty($TERMINAL)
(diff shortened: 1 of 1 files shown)


tpope:
`:Make` should never grab focus. You'll need to squash in my previous patch for starters.

zorab47:
I rebased my branch to include your commit. I can squash them together if you prefer.

Are there any other changes needed/

tpope:
Yes, please squash. After that it looks good to go, and I will personally test it.

zorab47:
I squashed them together. Thanks!

tpope:
Huh, this isn't consistently working for me because my terminal takes more than 100ms to launch sometimes. How about you?

tpope:
Found a solution and squashed it into your commit.

tpope:
BTW I am wondering if it would make more sense to minimize the window, except in the `:Start` case. Using a tiling window manager, I am finding the rearrange super disrupting.

zorab47:
With a tiling window manager that would be very disruptive. `wmctrl` doesn't offer a method of minimizing windows, at least nothing stood out. Some terminals may have a CLI switch for starting minimized, although `gnome-terminal` is missing the option.

[Searching more in depth][1] I found that a newer version of `wmctrl` does provide the iconify option, but it isn't currently available via apt on Ubuntu 14.04. The `xdotool` might be another option.

[1]:http://askubuntu.com/questions/488356/how-can-i-start-a-script-in-a-minimised-gnome-terminal

tpope:
Is there a difference between iconification and what `-b add,hidden` does?

zorab47:
Using `-b add,hidden` didn't hide or minimize the window when I tried it in Xfce.

tpope:
I guess no harm in adding it then :). Will investigate further in the
future.

PR #131, opened 2015-07-01
Use `echomsg`/`echoerr` for completion status

glittershark:
Sometimes I run a test suite, then check my phone or grab a coffee for
a couple minutes. If the suite passes, then when I come back I look at
vim and there's nothing there - so I second guess myself on whether I
actually ran the suite in the first place. Having some sort of
persistent message no matter what happens when a :Dispatch is run would
be nice.

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -789,11 +789,11 @@ function! dispatch#complete(file) abort
       let status = -1
     endtry
     if status > 0
-      echo 'Failed:' request.command
+      echoerr 'Failed:' request.command
     elseif status == 0
-      echo 'Succeeded:' request.command
+      echomsg 'Succeeded:' request.command
     else
-      echo 'Finished:' request.command
+      echomsg 'Finished:' request.command
     endif
     if !request.background
       call s:cgetfile(request, 0, 0)


glittershark:
Come to think of it, `echoerr` would be nice for a failure, too. I'll amend to include that

tpope:
Not sure I see the value here. `:echomsg` can disappear just as easily as `:echo`, no? The only difference is that you can get at the former by paging through to the very bottom of `:messages`, a task that gets increasingly arduous the longer Vim has been running.

Also, if my test suite fails, that's not a Vim error, and I would say shouldn't get an `:echoerr`. Plus, it should be obvious, since the quickfix pops open?

tpope:
I am thinking the underlying problem here might be alleviated with a timestamp? It's easy enough to capture but I'm not sure how to present it.

glittershark:
No matter what I do I don't get *any* output at all for a passing test suite after its completion. I don't know why specifically, but the `echo` just goes away completely. At least with echom there's *something* there

tpope:
Tmux?

glittershark:
Yep.

tpope:
Yeah, so I can't get the message to stick around in tmux no matter what I do. The tmux pane resize blows it away every single time. This change doesn't help.

glittershark:
Damn, okay.

tpope:
Another thing you can fiddle with is moving the `redraw`, adding a bang, or dropping it entirely. Or perhaps fiddling with `'lazyredraw'`. Since there is no downside, I would accept a patch of that nature even if it doesn't work for me.

It's really strange that `echomsg` works for you; I've made heavy use of it for debugging over the years but only found it helpful with `:messages`.

glittershark:
:+1: I'll fiddle around a bit

PR #138, opened 2015-08-03
Allow default error format to be customised

sharplet:
I'm personally using this to change the error format to `'%f:%l: %m,%+I%.%#'`. Currently output like this isn't parsed:

```
path/to/file.text:42: TODO: fix stuff
```

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -539,7 +539,7 @@ function! dispatch#compile_command(bang, args, count) abort
   call extend(request, {
         \ 'action': 'make',
         \ 'background': a:bang,
-        \ 'format': '%+I%.%#'
+        \ 'format': g:dispatch_format
         \ }, 'keep')
 
   if executable ==# '_'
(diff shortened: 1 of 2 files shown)


ericboehs:
This seems like a very necessary change. I'm not sure how one format was supposed to work. @tpope could we get this merged?

tpope:
Create a compiler plugin for the command you want to change the format for. Customizing the global fallback is a hack at best; it might fix one scenario but will inevitably break others.

ericboehs:
Hmm I'm not sure how to do that. I'm just trying to get rspec to work but the compiler plugin in Rails.vim isn't working. The matching syntax is `\v^([^:]+):([0-9]+).+`.

The first capture group being the file name and the second being the line. Is there documentation on how to add this?

tpope:
Vim ships with `compiler/rspec.vim` so that should work out of the box. If you are calling `bundle exec rspec` you might need to install [bundler.vim](/tpope/vim-bundler).

Issue #135, opened 2015-07-23
Fire user autocommands after dispatch is complete


wincent:
I've got another plug-in which I'd like to use to extend vim-dispatch's functionality. What do you think of having vim-dispatch fire an autocommand after each dispatch is complete, so that the other plug-in can hook in and do some extra stuff? (Specifically, I'm running searches and I want to do some post-processing over the results).

If you're open to the autocommands, let me know and I'll put together a pull request.

tpope:
It already fires a `cgetfile` when the results are loaded, which sounds like the event you want.

wincent:
Excellent, thanks @tpope.

PR #139, opened 2015-08-04
[RFC] Emit user event after opening quickfix window

mhinz:
Hi,

I found a few `:doautocmd` in the code, but they all fire before the quickfix window is opened. For other plugins that use dispatch and handle the quickfix window theirselves, it would be nice to have something like this.

What do you think?

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -854,6 +854,7 @@ function! s:open_quickfix(request, copen) abort
   if &buftype ==# 'quickfix' && !was_qf && a:copen != 1
     wincmd p
   endif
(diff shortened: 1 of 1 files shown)


justinmk:
See https://github.com/tpope/vim-dispatch/issues/135 . I think the relevant event is `QuickFixCmdPost`, though I am not clear on whether that happens before or after the window is opened.

mhinz:
_Before_, that's the issue. ;]

tpope:
Any reason you can't use `FileType qf`?

mhinz:
\*facepalms\*

I somehow didn't think of this. Thanks and sorry for the noise.

PR #143, opened 2015-08-12
Add capability to configure the tmux pane height

gabriel128:
Enable the chance of using g:dispatch_pane_size to configure the height of the pane opened in tmux
- Add extra conditional to tmux.vim file

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -850,7 +850,12 @@ endfunction
 
 function! s:open_quickfix(request, copen) abort
   let was_qf = &buftype ==# 'quickfix'
-  execute 'botright' (a:copen ? 'copen' : 'cwindow')
+  if exists('g:dispatch_quickfix_height')
+    let height = get(g:, 'dispatch_quickfix_height', 10)
+    execute 'botright' (a:copen ? 'copen' . height : 'cwindow' . height)
+  else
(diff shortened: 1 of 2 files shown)


phongnh:
+1

zabolotnov87:
+1

tpope:
Okay I guess I'll capitulate, but make it `get(g:, 'dispatch_quickfix_height', 10)`, and have it affect the quickfix window too.

Smeagol07:
+1
@tpope 
Can the quickfix use a separate setting? Like dispatch_quickfix_height and dispatch_tmux_height?
I would like to combine this option with ack.vim plugin and it would be nice to have small tmux panel (as one don't need to see a progress) but normal quickfix (as the final result is important).
I will make a pull request if that's ok.

Smeagol07:
+1
@tpope 
Can the quckfix have diferent option than tmux panel? Like `dispatch_quickfix_height` and `dispatch_tmux_height`?
I would like to use it in `ack.vim` plugin and it would be nice to have small panel when a search is in progress and nice quickfix at the end.
If so I will make a pull request.

gabriel128:
I'm going to close this PR since I moved to emacs, and I'd rather not write any more vimscript in my whole life :wave:

PR #157, opened 2015-12-13
Support for legacy %< format in makeprg string

jasonchoimtt:
Although %< exists for backward compatibility reasons (see :help %<), it would be nice to support it as well.


tpope:
Okay, but can you move it to the very beginning `<\=' so there's no false positive on `%:h<`?

jasonchoimtt:
Wow, I didn't think of these cases. Does it look good now?

tpope:
Yep

Issue #119, opened 2015-04-01
Output different when using tmux, related to shift in/out characters


shmup:
I'm not sure why tmux handles shift out/in characters (particularly with Dispatch) differently, but maybe it's something with my tmux configuration?

Also when I'm not using tmux I properly see the "Finished: ____" result.

The bottom example is tmux.

![output](https://cloud.githubusercontent.com/assets/118710/6954577/473754a4-d8a1-11e4-8cd7-5620f6845330.png)

tpope:
It's because the tmux adapter intercepts the pane output directly rather than piping through `tee`. Are those shift in/out characters supposed to do anything? You could try stripping them out in the tmux handler (the same way we currently do color codes and `\r`).

The `Finished` notification gets lost in the resize. Not a big deal because the resize itself is noticable.

PR #158, opened 2015-12-20
Strip `^O` from the output

teoljungberg:
References #119

### DIFF
--- a/autoload/dispatch/tmux.vim
+++ b/autoload/dispatch/tmux.vim
@@ -58,7 +58,12 @@ function! dispatch#tmux#make(request) abort
   elseif uname ==# 'Linux'
     let filter .= ' -u'
   endif
-  let filter .= " -e \"s/\r$//\" -e \"s/.*\r//\" -e \"s/\e\\[K//g\" -e \"s/.*\e\\[2K\e\\[0G//g\" -e \"s/\e\\[[0-9;]*m//g\" > ".a:request.file
+  let filter .= " -e \"s/\r$//\" -e \"s/.*\r//\""
+  let filter .= " -e \"s/\e\\[K//g\" "
+  let filter .= " -e \"s/.*\e\\[2K\e\\[0G//g\""
+  let filter .= " -e \"s/\e\\[[0-9;]*m//g\""
+  let filter .= " -e \"s/\017//g\""
+  let filter .= " > " . a:request.file . ""
   call system('tmux ' . cmd . '|tee ' . s:make_pane .
         \ (pipepane ? '|xargs -I {} tmux pipe-pane -t {} '.shellescape(filter) : ''))
 


autoload/dispatch/tmux.vim
```diff
@@ -58,7 +58,7 @@ function! dispatch#tmux#make(request) abort
   elseif uname ==# 'Linux'
     let filter .= ' -u'
   endif
-  let filter .= " -e \"s/\r$//\" -e \"s/.*\r//\" -e \"s/\e\\[K//g\" -e \"s/.*\e\\[2K\e\\[0G//g\" -e \"s/\e\\[[0-9;]*m//g\" > ".a:request.file
+  let filter .= " -e \"s/\r$//\" -e \"s/.*\r//\" -e \"s/\e\\[K//g\" -e \"s/.*\e\\[2K\e\\[0G//g\" -e \"s/\e\\[[0-9;]*m//g\" -e \"s///g\" > ".a:request.file
```
teoljungberg: Is there a better way of accomplishing this in `sed`?

tpope: No, but you can definitely do better than putting a binary control character in a Vim script.

tpope:
Why?

autoload/dispatch/tmux.vim
teoljungberg: I was skeptical over that too, could you give an example how this can be solved?

teoljungberg:
The rust compiler outputs `^O`, which are visible when using the tmux adapter.
This change strips those characters out before reading the output into vim's quickfix window.

Screenshots to demonstrate this are posted in #119

autoload/dispatch/tmux.vim
tpope: Like most other languages, VimL strings support octal escapes.

tpope:
That's more of a what than a why. Best I can tell it's a no-op to disable a mode that isn't enabled: https://en.wikipedia.org/wiki/Shift_Out_and_Shift_In_characters. So yeah, a good candidate to strip out.

teoljungberg:
@tpope I updated the PR to use `nr2char()` to remove the octal character.
I also refactored the variable `filter` to be split up into multiple lines, which might or might not be something we want to do.

Let me know if there's anything else I can do.

tpope:
Use an octal escape sequence, and squash.

teoljungberg:
Updated and squashed

autoload/dispatch/tmux.vim
```diff
@@ -58,7 +58,12 @@ function! dispatch#tmux#make(request) abort
   elseif uname ==# 'Linux'
     let filter .= ' -u'
   endif
-  let filter .= " -e \"s/\r$//\" -e \"s/.*\r//\" -e \"s/\e\\[K//g\" -e \"s/.*\e\\[2K\e\\[0G//g\" -e \"s/\e\\[[0-9;]*m//g\" > ".a:request.file
+  let filter .= " -e \"s/\r$//\" -e \"s/.*\r//\""
+  let filter .= " -e \"s/\e\\[K//g\" "
+  let filter .= " -e \"s/.*\e\\[2K\e\\[0G//g\""
+  let filter .= " -e \"s/\e\\[[0-9;]*m//g\""
+  let filter .= " -e \"s/\<C-O>//g\""
```
justinmk: `\017` is the octal

teoljungberg: Good call!
I found that this reads clearer, since it shows more explicitly which key we are stripping out.

What do you think?

tpope: It's "shift in", though, not somebody banging Control+O on a keyboard.

teoljungberg: Updated to use the octal escape now.

Issue #166, opened 2016-03-25
Displaying the executed command might cause hit-Enter prompt


blueyed:
The code at https://github.com/tpope/vim-dispatch/blob/eb3e564fb1edfee09fe6aeb57f3388e273da72c8/autoload/dispatch.vim#L218-L219 might trigger a hit-enter prompt (when it's too long / in a small window).

This causes issues, e.g. https://github.com/radenling/vim-dispatch-neovim/pull/4.
But even after fixing this, it seems like there's an issue where the tempfile from vim-dispatch-neovim only contains 10 empty lines, probably because the terminal window is still considered to be empty at that point (because of the hit-enter).

What do you think about only doing this `echo` if `&verbose` is > 0?

Is there some other way to handle this better?
As a workaround I could imagine cropping the line to something that fits into `cmdheight`, but that then seems to be more an issue to tackle at the core / vim.

/cc @radenling

tpope:
I think your workaround is the way to go: Truncate `a:request.expanded` such that the entire line fits within `&cmdheight * &columns`.

blueyed:
Given the fix for https://github.com/radenling/vim-dispatch-neovim/pull/4 I could imagine that it's a similar thing, where some commands fail with "not allowed here" or something similar, and could be replaced maybe?

btw: the "Success:" message simply does not get displayed if the window is too small - instead of the hit-enter prompt?! (https://github.com/tpope/vim-dispatch/blob/83e68d9fb1d60e8cdefe1e16e61b5dbf7d4f5680/autoload/dispatch.vim#L808)

After all, this should happen with `echo` in general, no?

blueyed:
> After all, this should happen with `echo` in general, no?

It works for displaying the cmd also when `redraw` is used after it.

`dispatch#complete` is called by dispatch-neovim in the `on_exit` handler, which appears to make Vim truncate the message automatically (with `T` in `'shortmess'`).
It would be nice if this mechanism could be triggered by vim-dispatch also for the case at hand.

blueyed:
A shorter/simpler approach would be to just call `:redraw` after the `echo` (which might be changed to `echom` then): this would avoid the hit-enter prompt and just not display the message.

But still: it would be better to find a to trigger Vim's auto-truncation (in the middle of the string).

tpope:
What constitutes "middle", even?

blueyed:
The middle of the string..?!

I agree that it's better to have control over this as with my PR.

[...]

PR #167, opened 2016-03-25
s:dispatch: shorten displayed message to avoid hit-enter prompt

blueyed:
Fixes https://github.com/tpope/vim-dispatch/issues/166.

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -375,12 +375,14 @@ function! s:dispatch(request) abort
       redraw
       let msg = ':!'
       let suffix = s:postfix(a:request)
+      let suffix_len = len(substitute(suffix, '.', '.', 'g'))
       let cmd = a:request.expanded
+      let cmd_len = len(substitute(cmd, '.', '.', 'g'))
       " NOTE: the extra "-13" is required to avoid the hit-enter, although
       " it's displayed on a single line already?!
       let max_cmd_len = (&cmdheight * &columns)-2-len(suffix)-13
-      if len(cmd) > max_cmd_len
-        let msg .= cmd[0:max_cmd_len-2] . '…'
+      if cmd_len > max_cmd_len
+        let msg .= '<' . matchstr(cmd, '\v.{'.(max_cmd_len - 1).'}$')
       else
         let msg .= cmd
       endif
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -381,7 +381,7 @@ function! s:dispatch(request) abort
       let max_cmd_len = (&cmdheight * &columns) - 2 - suffix_len - 2
 
       if has('cmdline_info')
-        let last_has_status = (&statusline == 2 || (&statusline == 1 && winnr('$') != 1))
+        let last_has_status = (&laststatus == 2 || (&laststatus == 1 && winnr('$') != 1))
 
         if &ruler && !last_has_status
           if empty(&rulerformat)
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -376,9 +376,9 @@ function! s:dispatch(request) abort
       let msg = ':!'
       let suffix = s:postfix(a:request)
       let cmd = a:request.expanded
-      " NOTE: the extra "-12" is required to avoid the hit-enter, although
+      " NOTE: the extra "-13" is required to avoid the hit-enter, although
       " it's displayed on a single line already?!
-      let max_cmd_len = (&cmdheight * &columns)-2-len(suffix)-12
+      let max_cmd_len = (&cmdheight * &columns)-2-len(suffix)-13
       if len(cmd) > max_cmd_len
         let msg .= cmd[0:max_cmd_len-2] . '…'
       else
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -372,15 +372,24 @@ function! s:dispatch(request) abort
     let response = call('dispatch#'.handler.'#handle', [a:request])
     if !empty(response)
       let a:request.handler = handler
+
+      " Display command, avoiding hit-enter prompt.
       redraw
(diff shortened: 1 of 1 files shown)


blueyed:
See https://github.com/vim-scripts/ingo-library/blob/558132e2221db3af26dc2f2c6756d092d48a459f/autoload/ingo/avoidprompt.vim#L31 for a more evolved method to get the maxlength correctly (instead of just `-12`).

autoload/dispatch.vim
```diff
@@ -215,8 +215,21 @@ function! s:dispatch(request) abort
...
       let pid = dispatch#pid(a:request)
-      echo ':!'.a:request.expanded .
-            \ ' ('.handler.'/'.(!empty(pid) ? pid : '?').')'
+
+      " Display cmd with handler and pid, avoiding a hit-enter prompt.
+      let msg = ':!'
+      let suffix = ' ('.handler.'/'.(!empty(pid) ? pid : '?').')'
+      let cmd = a:request.expanded
+      " NOTE: the extra "-12" is required to avoid the hit-enter, although
+      " it's displayed on a single line already?!
```
blueyed: More evolved: https://github.com/vim-scripts/ingo-library/blob/558132e2221db3af26dc2f2c6756d092d48a459f/autoload/ingo/avoidprompt.vim#L31

blueyed:
This should also be done in `dispatch#complete` then.

blueyed:
What about the suggestion from https://github.com/tpope/vim-dispatch/issues/166#issuecomment-201405421:

> A shorter/simpler approach would be to just call `:redraw` after the `:echom` (instead of `:echo`).

This would display it normally, but skip it in case there would have been a hit-enter prompt.
The benefit here would be that it's in `:mess`, in case you want to get to it later.

petobens:
Hi @tpope and @blueyed; I just bumped into this issue and was wondering if there is a chance to move this pr forward (with the corresponding necessary changes). Thanks in advance!

simlei:
Would love to see this PR or a similar change in vim-dispatch.

blueyed:
Rebased (old head: 0722f74).

autoload/dispatch.vim
```diff
@@ -373,7 +373,19 @@ function! s:dispatch(request) abort
...
       let a:request.handler = handler
       redraw
-      echo ':!'.a:request.expanded s:postfix(a:request)
+      let msg = ':!'
+      let suffix = s:postfix(a:request)
+      let cmd = a:request.expanded
+      " NOTE: the extra "-13" is required to avoid the hit-enter, although
+      " it's displayed on a single line already?!
+      let max_cmd_len = (&cmdheight * &columns)-2-len(suffix)-13
+      if len(cmd) > max_cmd_len
```
tpope: `len()` is byte length, it only works for ASCII. Use `len(substitute(cmd, '.', '.', 'g'))` to get the number of characters.

Maybe this explains the 12/13 discrepancy?

autoload/dispatch.vim
```diff
@@ -373,7 +373,19 @@ function! s:dispatch(request) abort
...
+        let msg .= cmd[0:max_cmd_len-2] . '…'
```
tpope: Use `'<' . matchstr(cmd, '\{' . max_cmd_length - 1 . '\}$')` to get the correct length and match Vim's usual truncation (avoiding any encoding issues from the ellipsis in the process).

blueyed:
> Maybe this explains the 12/13 discrepancy?

No.
But with Vim I only need `-2` (which makes more sense)!  (I was using Neovim apparently already when reporting the issue in 2016)
This is the case with Vim 7.4.2200 already (the oldest to really support async (#283)).
But have checked it using `echo repeat('X', &columns-1)` back to Vim 7.4.41.

Requiring more there is a bug in Neovim, and this patch here should maybe cope for that, but I'll look into getting it fixed there also.

blueyed:
The difference is due to (not) using `set showcmd` etc.
Something like this would be useful: https://github.com/Shougo/unite.vim/blob/e1d3bb9f836a8645807b790345b69666bf22bb2f/autoload/unite/view.vim#L799-L826

blueyed:
Even with (re)setting showcmd it will overwrite the tail when (re)setting it.
I've filed an issue for Vim itself: https://github.com/neovim/neovim/issues/10601.

But until it is fixed (and for older versions) I think we should check the two settings (ruler, showcmd) and take those into account for the max length.

autoload/dispatch.vim
```diff
@@ -375,12 +375,14 @@ function! s:dispatch(request) abort
...
+      let suffix_len = len(substitute(suffix, '.', '.', 'g'))
       let cmd = a:request.expanded
+      let cmd_len = len(substitute(cmd, '.', '.', 'g'))
       " NOTE: the extra "-13" is required to avoid the hit-enter, although
       " it's displayed on a single line already?!
       let max_cmd_len = (&cmdheight * &columns)-2-len(suffix)-13
-      if len(cmd) > max_cmd_len
-        let msg .= cmd[0:max_cmd_len-2] . '…'
+      if cmd_len > max_cmd_len
+        let msg .= '<' . matchstr(cmd, '\v.{'.(max_cmd_len - 1).'}$')
```
blueyed: Initially this would display the beginning, now it displays then end/tail.

tpope: I think that's fine, the most important part is this `[whatever/123]` at the end anyways.

blueyed:
For reference: `v:echospace` was added in Vim patch 8.1.1913 (https://github.com/vim/vim/commit/37f4cbd46f5a6f2dd3a48d5fa4324dce37e4bd6c).

petobens:
@blueyed will that allow to close https://github.com/tpope/vim-dispatch/issues/166?

blueyed:
@petobens 
Yes, it's meant to help with that.

PR #174, opened 2016-06-10
iTerm 3 compatability

scizo:
iTerm 3 introduced breaking changes to the AppleScript api, mentioned
here https://www.iterm2.com/version3.html and here
https://www.iterm2.com/documentation-scripting.html. This allows the
iTerm dispatcher to work with either the version 2 or version 3 api. I wasn't
sure how activate should be triggered, so I am not entirely sure that the version 3
activate script is working, but would be happy to test it with a little guidance.


z0rc:
@tpope do you have any update on this?

Issue #160, opened 2015-12-26
Applescript would throw error when works with iTerm2 > 2.9


chemzqm:
The reason is iTerm2 changed it's applescript API: https://iterm2.com/applescript.html
The perl script runs in bash would also not works.

I've tried to make a eay fix for that https://gist.github.com/chemzqm/499003e3f21772397986 in case someone need it.  It's using some fish script as I use fish shell for my default profile.

petobens:
@tpope dispatch indeed doesn't seem to work with latest (test or nightly) release of iTerm2. Is it possible to fix it? Thanks!!1

PR #176, opened 2016-07-15
Fix for new iTerm2 API

fanzeyi:
Tested with ITerm2 Build 3.0.4.

Based on the work of @chemzqm in issue #160. I made some changes on using the original script generated by Dispatch.vim rather than the fish script. So Dispatch.vim could receive the completion message correctly.


fanzeyi:
Actually, I found #174 has better a compatibility. So I'm closing this pull request. Sorry for bothering.

PR #186, opened 2016-09-15
Add User#DispatchQuickfix, set t:dispatch_quickfix_winnr

blueyed:
In `s:open_quickfix` a user autocommand is triggered after the quickfix
window has been massaged, which includes setting `w:quickfix_title`.
This has to be done in the end, since it is not available with `au
FileType qf`.

The variable `t:dispatch_quickfix_winnr` is set to inform the user about the
window's number.

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -867,6 +867,16 @@ function! s:open_quickfix(request, copen) abort
       if has_key(a:request, 'compiler')
         call setbufvar(bufnr, 'current_compiler', a:request.compiler)
       endif
+      let t:dispatch_quickfix_winnr = winnr
+      if exists('#User#DispatchQuickfix')
+        try
+          let [save_mls, &modelines] = [&mls, 0]
+          doautocmd User DispatchQuickfix
+        finally
+          let &mls = save_mls
+        endtry
+      endif
+      break
     endif
   endfor
 endfunction


tpope:
What's your use case? I'm increasingly of the mind that `User` autocommands are a gross kludge, and I'd rather jam this into `QuickFixCmdPost` or something if at all possible.

blueyed:
My use case is to adjust sizes for (all) quickfix windows (https://github.com/blueyed/vim-qf_resize/blob/de23eb05e0eb569e152cba37c5f1676d19dc5420/plugin/qf_resize.vim#L223).  It handles dispatch windows in a special way, and needs to adjust based on that.
If `QuickFixCmdPost` gets run after the window has been opened, this might be fine, but IIRC it did not work.

tpope:
Okay, `QuickFixCmdPost` won't work. We could trigger `FileType` again, if that would help. But I'm still confused. You say you need to handle dispatch windows specially, but it looks like you found a way to do that in checking `w:quickfix_title` *before* dispatch.vim clobbers it. So why do you need to do it afterwards?

I don't see addressing this until after some additional changes to title behavior (taking advantage of Vim 8's `setqflist({})`). But if I understand your exact needs I can maybe knock both things out in one go.

blueyed:
I think my use case is covered actually via `FileType` already, where I can look at `w:quickfix_title`, which starts with `:Dispatch`.
Also `b:dispatch` is set in the qf window.

Closing, thanks!

PR #187, opened 2016-09-15
Call s:open_quickfix in the context of the request's dir

blueyed:
This makes the paths relative to where the compiler was invoked from.
Without this, the quickfix window will use the working dir from the
current window, which is often not the case when using `:FocusDispatch` etc.

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -824,27 +824,30 @@ function! s:cgetfile(request, all, copen) abort
   let dir = getcwd()
   let modelines = &modelines
   try
-    let &modelines = 0
-    call s:set_current_compiler(get(request, 'compiler', ''))
-    exe cd fnameescape(request.directory)
-    if a:all
-      let &l:efm = '%+G%.%#'
-    else
-      let &l:efm = request.format
-    endif
-    let &l:makeprg = request.command
-    silent doautocmd QuickFixCmdPre cgetfile
-    execute 'noautocmd cgetfile' fnameescape(request.file)
-    silent doautocmd QuickFixCmdPost cgetfile
-    call s:open_quickfix(request, a:copen)
-  catch '^E40:'
-    return v:exception
+    try
+      let &modelines = 0
+      call s:set_current_compiler(get(request, 'compiler', ''))
+      exe cd fnameescape(request.directory)
+      if a:all
+        let &l:efm = '%+G%.%#'
+      else
+        let &l:efm = request.format
+      endif
+      let &l:makeprg = request.command
+      silent doautocmd QuickFixCmdPre cgetfile
+      execute 'noautocmd cgetfile' fnameescape(request.file)
+      silent doautocmd QuickFixCmdPost cgetfile
+    catch '^E40:'
+      return v:exception
+    finally
+      let &modelines = modelines
+      let &l:efm = efm
+      let &l:makeprg = makeprg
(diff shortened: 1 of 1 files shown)


tpope:
This could cause the restoration code in the finally clause to erroneously
run in the quickfix window.

blueyed:
Another `try`/`finally` for the `cd` then?

tpope:
I guess put the second `try`/`finally` in `s:open_quickfix()`? You'll need to make sure the quickfix window is focused, which will add complexity.

blueyed:
@tpope 
What do you think about the fixup?  (I've pushed it before seeing your comment)

In the same area: it would be nice to know if `:Copen!` was used, but setting a buffer local var for this will keep it with `:colder` then, too.
Maybe that should be provided by/via https://github.com/tpope/vim-dispatch/pull/186 instead.

blueyed:
I have a more reliable way to handle the quickfix window by now, using
Vim's new window id feature and the method to query the qf list:

```vim
  try
    let qfinfo = getqflist({'winid': 1})
    let winnr = qfinfo['winid']
  catch /^Vim\%((\a\+)\)\=:E118/
    echom "Checking" winnr('$') "windows"
    for winnr in range(1, winnr('$'))
      if getwinvar(winnr, '&buftype') ==# 'quickfix'
        if len(getloclist(winnr))
          " This is a location list (or a qf list with a location list
          " (unlikely)).
          continue
        endif
        break
      endif
    endfor
  endtry
```

I am using a mix of this PR and #186 locally since a while.

Please let me know if I should merge them into a new PR etc.

tpope:
Let me put it a different way. What happens when `haslocaldir()` is false? No amount of rearranging `s:cgetfile()` is going to cause an `lcd` in the quickfix. You *need* to address this in `s:open_quickfix`.

I'm pushing my own fix to the `quickfix-cd` branch. It addresses my concerns but might have other issues. Please try it and report back.

tpope:
Regarding the `:Copen!` check, maybe we should be setting `&l:efm` to reflect that? It has the same `:colder` problem, but I don't think there's any way around that.

tpope:
Closing because I accidentally pushed to master, but the solicitation for feedback still applies.

blueyed:
With 605631c it might still get to an unrelated (e.g. location list) window (https://github.com/tpope/vim-dispatch/pull/187#issuecomment-250146635).

Re `:Colder!` I'd prefer a buffer-local var, but `&l:efm` makes sense regardless (if it's not done already?!).

`getqflist({'all': 1}).nr` seems to be useful here: if `:Colder!` is used, a buffer local var could be set to this number.  It requires a recent Vim, but that seems to be OK.

tpope:
Good call on the location window. Ported your fix over.

Per usual, you're going to have to make a really, really good case if you want me to commit to a whole new variable. Honestly even `:Copen!` has felt kind of superfluous ever since I started stripping `%-G%.%#` out of the error format. I don't want to keep digging that hole deeper.

We can do a lot better with `getqflist({'all': 1}).nr` than just yet another variable. We could track which dispatch request corresponds to which quickfix list and automatically update the variables each time the list changes.  Please do experiment, just keep backwards compatibility in mind.

tpope:
Bad news: `getqflist({'all': 1}).nr` isn't a persistent identifier. It's always between 1 and 10.

Good news: `.title` could work.

Issue #188, opened 2016-09-18
Adapter for background jobs


teoljungberg:
With the release of Vim 8, we got support for [background jobs](http://vimhelp.appspot.com/eval.txt.html#job_start%28%29). Which I can see as something I'd like to be incorporated into dispatch.vim.

I'd gladly work on this myself, but I wanted to test out the waters and ask prior to starting developing this as a pull request to dispatch.vim itself, or as a plugin in the likes of [vim-dispatch-neovim](https://github.com/radenling/vim-dispatch-neovim).
If so, could you provide any directions, tips and points, to get me started?

If this is not something we see as it could fit into the current goals of dispatch.vim, that's fine too.

tpope:
There's a lot of ways you could go with this. Do you want it to capture output and show it in a tab or something? My guess that would be convoluted and fiddly and probably something I would avoid merging until I'm sure the kinks are worked out. Running it invisibly, on the other hand, is probably pretty simple and a candidate for a quick merge, though probably a bit less useful.

myw:
From a cursory example, it looks like it would be possible to just add a `dispatch/jobs.vim` file, in which you wrap the existing dispatch commands around the new low-level `:job_` functions/commands in Vim 8. This would be really cool, as it would let existing plugins that depend on `vim-distpatch` transition to using the new built-in job functionality (gaining its perks in speed and reliability) without having to rewrite their own async-handling. Since I haven't actually tried it out, I don't know how many kinks there would be to iron out, but it seems _eminently_ doable, with a large upside. If no one takes this on, I'd look into it over a weekend project.

teoljungberg:
> There's a lot of ways you could go with this. Do you want it to capture output and show it in a tab or something? 

My ideal use case would be to have it work like the other adapters work, to have the errors be read into the quickfix list.

tpope:
Well the other adapters generally have a way to watch the command in progress, in a tmux pane or terminal window or whatever. If that isn't a requirement, it'll probably be pretty simple.

tpope:
Let's centralize discussion in #193.

Issue #28, opened 2013-04-11
Feature: Ability to cancel / abort a dispatched command


dhruvasagar:
Not sure how many would find this useful, but I would like to be able to abort a dispatched task and most importantly ignore it's output and not have it loaded into quickfix list.

Perhaps I should :Start such a command rather than :Dispatch, but in either case currently there's no support for being able to abort it from within vim and as of now I am not even sure if it is feasible, so just putting it out there, and I can see this to be useful in both cases with :Start and :Dispatch.

bounceme:
Something like this would be helpful in stopping long running processes that are started headless.

tpope:
Went with a simple string match for aborting older tasks. I could see adding something similar to `:Copen`, to help with @shym style scenarios. This doesn't preclude adding other specification methods, though I think the `@^` style stuff is probably overkill.

PR #193, opened 2016-10-24
Add initial vim jobs strategy

edkolev:
This is an initial version of a strategy based on vim 8's jobs. I saw the discussion in #188, hence this PR.

I don't think it's ready yet (I haven't even tested it on files with whitespace in them).

This strategy works only for 'make' actions, but maybe it should also have a check if 'background' is true.

Any feedback is appreciated.

Edit: I haven't thought about windows support at all - the strategy uses dispatch#prepare_make() as well as &shell and &shellcmdflag

### DIFF
--- a/autoload/dispatch/jobs.vim
+++ b/autoload/dispatch/jobs.vim
@@ -1,0 +1,63 @@
+" dispatch.vim jobs strategy
+
+if exists('g:autoloaded_dispatch_jobs')
+  finish
+endif
+let g:autoloaded_dispatch_jobs = 1
+
+let s:waiting = {}
+
+function! dispatch#jobs#handle(request) abort
+  if !(has('job') && has('channel') && a:request.action ==# 'make')
+    return 0
+  endif
+
+  let channel_id = s:start_make_job(a:request)
+  if empty(channel_id)
+    return 0
+  endif
+
+  let s:waiting[channel_id] = a:request
+  return 1
+endfunction
+
+function! s:start_make_job(request)
+  let command = dispatch#prepare_make(a:request)
+  let job_options = {
+        \ 'close_cb': function('s:CloseCallback'),
+        \ 'exit_cb' : function('s:ExitCallback'),
+        \ 'out_io'  : 'pipe',
+        \ 'err_io'  : 'out',
+        \ 'in_io'   : 'null',
+        \ 'out_mode': 'nl',
+        \ 'err_mode': 'nl'}
+  let job = job_start([&shell, &shellcmdflag, command], job_options)
+  return s:channel_id(job)
+endfunction
+
+function! s:channel_id(job)
+  if job_status(a:job) ==# 'fail'
+    return ''
+  endif
+
+  let channel = job_getchannel(a:job)
+  if string(channel) ==# 'channel fail'
+    return ''
+  endif
+
+  let channel_info = ch_info(channel)
+  return string(channel_info.id)
+endfun
+
+function! s:ExitCallback(job, status)
+  let channel_id = s:channel_id(a:job)
+  if has_key(s:waiting, channel_id)
+    let request = remove(s:waiting, channel_id)
+    call dispatch#complete(request)
+  endif
+endfun
+
+function! s:CloseCallback(channel)
+  " trigger vim calling s:ExitHandler()
+  call job_status(ch_getjob(a:channel))
+endfun
--- a/plugin/dispatch.vim
+++ b/plugin/dispatch.vim
@@ -35,6 +35,7 @@ if !exists('g:dispatch_handlers')
   let g:dispatch_handlers = [
         \ 'tmux',
         \ 'screen',
+        \ 'jobs',
         \ 'windows',
         \ 'iterm',
         \ 'x11',


z0rc:
What's the status of this PR? Is it going to be updated? Does it require review?

tpope:
My first reaction is that we almost certainly should not be using `dispatch#prepare_make()`. We can get the exit status from `exit_cb` and replace the redirect with `out_file`. I think that just leaves capturing the pid. Is there a clean way to do that? I see it in the string form of the job object. Perhaps we just munge it out of there?

blueyed:
Just for reference: vim-dispatch-neovim uses Neovim's terminal, and I've been using it for a while (with [some modifications](https://github.com/radenling/vim-dispatch-neovim/pull/11/files))

tpope:
I pushed my [own attempt](https://github.com/tpope/vim-dispatch/commit/14de90f309ca1aa3232592eaea6b4f7b28b3c6bd). It works but it's really unnerving not to see any output till it finishes.

shym:
@tpope I’ve been giving your job branch a try since I would love to see it merged in the main branch.
I noticed a difference with the tmux handler: the command is not interpreted by a shell, so `Dispatch echo a; echo b` yields `a; echo b` in job, and `a\nb` in tmux.

tpope:
Just forced pushed a bunch of local changes. Try again.

shym:
The progressive output in the quickfix window is great!

- It would be nice if the height of the window could be configured (I often use a small screen so I set dispatch_quickfix_height). I guess I would even prefer if there was a way to set a different (smaller in my case) height while the compilation happens (since that output is useless until at least an error is found).
- It would be nice also to have support for background compilations.
- There is a typo in `compiler` in `get(request, 'complier', '')`.

shym:
I may add that I’d love to be able to set dispatch_quickfix_height and the new dispatch_running_quickfix_height (?) to 0 in special cases when I want this to run without immediate visual feedback (when I run tools like style-checkers in autocmd but want to take into account their report only once in a while).

tpope:
Before we get into bikeshedding I have some more fundamental concerns. Jobs are really bad for commands that read [from stdin](https://github.com/erikhuda/thor/pull/563) or [from /dev/tty](https://github.com/tpope/vim-fugitive/issues/722). That's just 2 examples, but infinite loops and hangs are as bad of examples as you can get. Issues with /dev/tty are probably a lost cause, but maybe we can use channel support to handle the stdin case?

shym:
For `/dev/tty` vim seems to properly detach jobs from the terminal:
`:Dispatch printf 'url=http://whatev.er\n\n' | git credential fill`
yields, as hoped:
`|| fatal: could not read Username for 'http://whatev.er': No such device or address`
which is obviously not ideal but much preferable to a hang as in fugitive#722.

tpope:
You're right, I must have confused myself testing on Neovim or something.

shym:
For commands that break when stdin is closed, I’m tempted to say that they’re just buggy and should be fixed; but that’s not a very constructive answer. Only some simple workaround comes to mind (such adhoc commands to interact with the command), maybe because I don’t see many use cases for compilers dropping to interactive mode; the only case I know is latex, and so I just tried and seen that it then exits with a fatal error if its stdin is closed (which is a much nicer behaviour than its standard one IMHO).

tpope:
I agree, it would be nice to solve but we're down to basically one broken program so it's no longer a deal breaker. Other issues:

* Sometimes it fails to capture all output, presumably because the exit callback fires while output is still buffered.
* Incrementally updating the quickfix list won't work if the error format contains multi-line items. We should check the format and do a full re-parse if that's the case (which hopefully won't have performance problems).

shym:
Do you have an example for the issue about failing to capture all output, to try a way to address it?
I wrote a silly `talktoomuch.c` "compiler":

```c
#include <stdio.h>

int main(int argc, char *argv[]) {
    int i;

    for (i = 0; i < 10000; i++)
        printf("but yeah\nbut no\n");

    printf("/tmp/a.c:2:13: warning: something to say\n");

    return 0;
}
```

Running a `Dispatch -compiler=gcc /tmp/talktoomuch` lasts horribly long (about 15s on my slow computer), keeping one core fully busy while lines get added to the quickfix list. Dispatch reports a success immediately but the final error gets eventually added to the quickfix list and recognized as such.
`:h read-in-close-cb` makes me wonder if `s:output` could detect whether it reached the end of the output to process, so that it could invoke `s:exit` (and when `s:exit` is called by vim as the exit callback, it could then abort if there is still some output buffered).

To compare performance, I ran my simple `talktoomuch` with the tmux adapter. Obviously it performs really fast but every time I got a truncated output (so the actual error is never captured) :-/

tpope:
Unfortunately it's only happened when I've been too immersed to investigate. It's always something that fails fast, with the error message getting omitted. `:h read-in-close-cb` is a good find. Next time it happens I'll rig something up.

shym:
I realised that one source of slowness of `Dispatch -compiler=gcc /tmp/talktoomuch` was an autocommand that fired on every `caddexpr`. I replaced my `autocmd QuickFixCmdPost *` with something more fine-grain, to remove `caddexpr`.
Along with a full re-parse when needed, it might be relevant to trigger an autocmd when the whole thing (the process exited and its output was handled by vim) ends.
I still get the inefficiency from vim handling the output line by line, ever so slowly.

By the way, combining the fact that detecting when everything did really end, the fact that sometimes a full re-parse is needed anyway, I wonder if it might just be more efficient to `tee` a copy of the output to a file and `cgetfile` it as soon as the process exits, if it is possible to interrupt the `'callback'` from being called (I just tried hacking horrible things like a `remove` of the `callback` field in the `s:exit` function but even if it was horrible it didn’t work ;-).

gauteh:
I'm testing the job branch now and it seems to work fine. Is there a way to Cancel the running job, type Ctrl+C?

shym:
@gauteh: you might have a look at #28 for the discussion about such a feature (and / or the abort branch). Indeed, it would be more useful with the job adapter than with most others.

gauteh:
@shym: nice. that should do the trick, from what I could quickly glance it does not seem to abort Vim jobs yet?

tpope:
Took a quick look at this. The `:Abort` command essentially just calls `kill -HUP <pid>`, and it seems to be correctly doing that on the job branch as well. However, unlike other adapters, that use a wrapper script, jobs are invoked with a simple `bash -c 'your command'`, and bash is smart enough to change this `exec your command`, which means `your command` receives the `HUP` directly. Some commands may ignore `HUP`, causing the abort to appear to fail.

This is annoying. We could add a postfix like `; exit $?`, but now we have to worry about alternate shells like `fish` and `cmd.exe`.

gauteh:
Tim Pope writes on september 5, 2017 23:44:
> Took a quick look at this. The `:Abort` command essentially just calls `kill -HUP <pid>`, and it seems to be correctly doing that on the job branch as well. However, unlike other adapters, that use a wrapper script, jobs are invoked with a simple `bash -c 'your command'`, and bash is smart enough to change this `exec your command`, which means `your command` receives the `HUP` directly. Some commands may ignore `HUP`, causing the abort to appear to fail.
> 
> This is annoying. We could add a postfix like `; exit $?`, but now we have to worry about alternate shells like `fish` and `cmd.exe`.

Would it be possible to do:

`:Abort -9` to send the KILL signal in stead? I assume this will also be 
propagated to the underlying process. 

Alternatively, for this particular adapter `:Abort` could use: 
`job_stop(job)` or `job_stop(job, "kill")`. That would be the 
most vimsy way.

stevenharman:
*/me: apologizes in advance for the drive-by comment.*

Is adding a Vim 8 Jobs strategy still being considered? As a Vim-in-terminal (usually, MacVim in iTerm, but whatever) user I've long wanted async dispatch w/o the need to also have tmux or the like set up. The fewer moving parts, and all of that. Anyhow... just curious if there's still interest and/or movement on adding it.

Thanks for making so many of the tools we all take for granted. ❤️

tpope:
I have decided that the next step will be to merge in the job adapter and make it low precedence, and maybe provide an option to make it high precedence (since adjusting the list manually is a pain). First though, I'd like to add Neovim support. Anyone feel like knocking that out? I'm looking for direct to quickfix list support like the Vim 8 implementation, not `:terminal`.

tpope:
I was going to merge the job branch immediately after 1.6 and worry about Neovim later, but ran into a few issues:

* Sometimes the last few lines would get discarded because the exit event fired while there was still output queued. I've pushed some additional changes to the job branch that seem to resolve this, but it's a confusing belt-and-suspenders mess I don't have a lot of confidence in.
* If you use `:colder` and `:cnewer` to switch between quickfix lists, output can be lost. We need some sort of buffer.
* `:AbortDispatch` is not particularly effective at stopping a running task compared to CTRL-C. I think this is because we always signal the top most pid, rather than the foreground task. Maybe `job_stop()` would be more effective.

shym:
About the first issue, I was thinking it might be simpler by hooking on `close_cb` instead of `exit_cb`? Of course, we probably want to wait for the two events to have fired, to get the exit status and know that all the output was processed.
At the moment, the job strategy is pretty much unusable for me because my (really sloooow) disk is spinning like crazy while output lines are added one at a time, with half a second between each, even if the job finished long ago. Might it be the current tricks that cause it to be so slow?

About the third issue, I’m not sure I managed to notice the same slowness but maybe `kill -HUP -pid` might be more efficient to kill them all at once.

tpope:
I've added a buffer, `job_stop()` for aborting, and Neovim support. I think this is basically ready. Further testing appreciated.

@shym I'm not sure why it would be slow. I've added `close_cb` support so if it was related to the old "tricks" it might have cleared up.

teoljungberg:
@tpope The [job] branch doesn't seem to be updated with these changes - are they pushed elsewhere?

[job]: https://github.com/tpope/vim-dispatch/tree/job

tpope:
It’s definitely there, maybe the fact I force pushed misled you?

tpope:
This is *finally* on master. I decided to prioritize it just below tmux because that's only other adapter that rivals the experience.

simlei:
Thanks so much for your work. Updating my vim-dispatch today and have ordinary script output suddenly and without further config "just working" in a much improved fashion -- continuous updates in the quickfix list, without loss of focus -- really made my dev day :)
it's really a good way of improving a plug-in that you're doing here :)

PR #195, opened 2016-11-29
s/program/command in docs

JoshTGreenwood:
The terms 'program' and 'command' are intermingled in the docs.
This changes every reference to be 'command' which I feel more
adequately represents the goal of vim-dispatch


tpope:
The style here is that `{program}` corresponds to `'makeprg'` and is distinct from `[arguments]`, while `{command}` represents the whole shebang and is used where the distinction does not matter. If you have further suggestions in light of this, feel free to submit a new patch.

PR #207, opened 2017-07-22
Add a count to Copen to set quickfix height

### LINKED ISSUE #177: Add a height option to `Copen`
`copen` command can take a height option for the quickfix window and, as far as I can tell, that height is preserved until the next explicit invocation of `copen`.
Could `Copen` provide the same option and behaviour?

By the way, for consistency, it would be great for that height to be used also for tmux splits.

shym:
Proposal of a PR to close #177.
If you agree with this solution, I could mention the option in the documentation.


tpope:
I don't think I'm on board with the whole `dispatch_temporary_quickfix_height` shenanigans.

shym:
Thinking about it, I’m not so sure either.
So I rewrote this PR to what I originally had in mind, when I opened #177.
This does change a current behaviour for users who set `dispatch_quickfix_height` and expect `Copen` to use this value, but this makes it closer to `copen`, using 10 lines unless you explicitly mention the height.

shym:
@tpope did you have time to look that PR again?

PR #208, opened 2017-07-22
Propose a setting to disable automatic copen-ing

shym:
Propose to extend the meaningful settings for the height of quickfix window with 0, to state that it should not be opened automatically. This is inspired by what happens with `:make`. I find this especially useful when I’m short on screen space (which is quite often).
This PR consists of one real commit, but it’s based on #207 since it modifies related lines.

Edit: changed the commit number to a reference to the PR.

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -1091,6 +1091,9 @@ endfunction
 function! s:cwindow(request, all, copen) abort
   call s:cgetfile(a:request, a:all)
   let height = get(g:, 'dispatch_quickfix_height', 10)
+  if height <= 0
+    return
+  endif
   let was_qf = s:is_quickfix()
   execute 'botright' (a:copen ? 'copen' : 'cwindow') height
   if !was_qf && s:is_quickfix() && a:copen !=# -2


tpope:
I guess I'm okay with using a height of zero as an undocumented way to surpress opening the quickfix window. I'm less keen on adding a `-count` to `:Copen`, as I'm thinking I might use the argument to open older commands, at some point.

shym:
Great. So I rebased the commit of this PR on master.
(After quite some time of using `-count` to set the height, I didn’t find it as convenient as I hoped anyway.)

tpope:
So I initially took this to only affect automatic opening. Having a manual call to `:Copen` not actually `:copen` is kind of weird. I guess I'm going to merge for now as an unofficial, undocumented behavior but will probably revisit at some point. Maybe I'll add `:Cgetfile`, or something. Vim 8's quickfix enhancements like `:chistory` have me rethinking how all this should work.

Side note, in #177 I see I said I'd accept `-count`. Guess I'm more fickle than even I realized.

shym:
Great! And I agree that something along the lines of `Cgetfile` would be more elegant.

PR #209, opened 2017-07-22
Propose a setting to run tmux foreground jobs in a separate window when possible

shym:
Propose to extend the meaningful settings for the height of the tmux split for foreground jobs with 0, to state that it should be run in a different (background) window when the GUI subsystem is enabled (so that the job will be able to call back when it finishes).
With this setting, one can obtain the same behaviour with tmux than with screen.
I find this especially convenient whenever I use compilers which takes a long time to start and output eventual messages in one go at the end.

This PR consists of one real commit, but it’s based on #208 since it modifies related lines.
I’d be happy to reorganise (and document) all this if there’s a better way.

Edit: changed the commit number to a reference to the PR.

### DIFF
--- a/autoload/dispatch/tmux.vim
+++ b/autoload/dispatch/tmux.vim
@@ -49,12 +49,12 @@ function! dispatch#tmux#make(request) abort
 
   let title = shellescape(get(a:request, 'title', get(a:request, 'compiler', 'make')))
   let height = get(g:, 'dispatch_tmux_height', get(g:, 'dispatch_quickfix_height', 10))
-  if get(a:request, 'background', 0)
+  if get(a:request, 'background', 0) || (height <= 0 && dispatch#has_callback())
     let cmd = 'new-window -d -n '.title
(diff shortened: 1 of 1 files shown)


tpope:
So if height is zero and `dispatch#has_callback()` is false, it just ... breaks? In the not too distant future I will make `dispatch#has_callback()` mandatory but until then I want to handle this case gracefully.

shym:
In all my tests, `tmux split -l 0` (or 1, for that matter) creates a 2-line split, just like `tmux split -l 2`. It’s been like this since tmux v2.5 at least, so I didn’t force the height to another value when `dispatch#has_callback()` is false. You think we should?
If the height is negative, it breaks, so it might be a reason to fix it (even if setting a negative height is a bit weird).

tpope:
Oooh how about `(height < 0 ? -height : height)` so you can provide your own fallback. I love dumb clever shit like that.

shym:
Nice ;-) I updated.

tpope:
`abs()` was added in 7.3, let's stick to the ternary.

shym:
Ok, done.

PR #223, opened 2017-12-07
X11: pass command as single argument to terminal

benburrill:
In some terminal emulators, -e takes only a single argument.  In
particular, I ran into problems with the xfce4-terminal; xterm allows
either form.

Testing:
TERMINAL=xfce4-terminal vim +"Spawn yes it works" +q
TERMINAL=xterm vim +"Spawn yes it works" +q

Edit:
In retrospect, I'm not sure this is the best option, as further testing shows that st and the rxvt terminal emulators do not accept a full command line as a single argument. However, xfce4-terminal is not unique in the way it treats -e either: gnome-terminal does the same thing.

Dropping st and (u)rxvt support seems unwise, but for now I'm going to leave this PR open because it would be nice if vim-dispatch worked with my `$TERMINAL`, I just don't think this is the right way to do it.

### DIFF
--- a/autoload/dispatch/x11.vim
+++ b/autoload/dispatch/x11.vim
@@ -39,7 +39,8 @@ function! dispatch#x11#spawn(terminal, command, request) abort
   if a:request.background || a:request.action ==# 'make'
     let command = 'wmctrl -i -a '.v:windowid . ';' . command
   endif
-  call system(dispatch#shellescape(a:terminal, '-e', &shell, &shellcmdflag, command). ' &')
+  let command = dispatch#shellescape(&shell, &shellcmdflag, command)
+  call system(dispatch#shellescape(a:terminal, '-e', command) . ' &')
   return 1
 endfunction
 


benburrill:
In retrospect, I'm not sure this is the best option, as further testing shows that rxvt terminal emulators do not accept a full command line as a single argument.  However, xfce4-terminal is not unique in the way it treats `-c` either: gnome-terminal does the same thing.

For now I'm going to leave this PR open because it would be nice if vim-dispatch worked with my `$TERMINAL`, but this is maybe not the right way to do it.

tpope:
Well that's annoying. What if we only add `-e` if `$TERMINAL !~# '-.$'`? That would allow `TERMINAL="xfce4-terminal -x"` to work, along with `gnome-terminal --` (as `-e` and `-x` appear to both be deprecated).

benburrill:
I think it might just be better to have a vim variable that overrides whatever `$TERMINAL` is set to.  Then people can configure dispatch to jibe with whatever they're using.

`$TERMINAL` is used in other contexts, and having `TERMINAL="xfce4-terminal -x"` can be problematic.  For the purpose of simply launching an new terminal window with `$TERMINAL`, it would not work, because `xfce4-terminal -x` requires that command to be specified.  It also causes other tools like `i3-sensible-terminal` to use a fallback terminal and not the xfce4 one.

tpope:
I was afraid there might be something like that. We'll go with an option for now. I'd be open to adding some smarter checks on `$TERMINAL`, and maybe even a query for the default Gnome/XFCE terminal (since setting an environment variable can be awkward if you aren't using `.xsession`).

PR #230, opened 2018-03-14
Set global focus to `:.Dispatch` with `:.Focus!`

teoljungberg:
From the following discussion: https://github.com/tpope/vim-rails/issues/506#issuecomment-358405491

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -706,10 +706,15 @@ function! dispatch#focus_command(bang, args, count) abort
   if has_key(opts, 'directory')
     let args = dispatch#dir_opt(opts.directory) . args
   endif
-  if empty(a:args) && a:bang
+  if empty(a:args) && a:bang && a:count < 0
     unlet! w:dispatch t:dispatch g:dispatch
     let [what, why] = dispatch#focus(a:count)
     echo 'Reverted default to ' . what
+  elseif empty(a:args) && a:bang && a:count > 0
+    unlet! w:dispatch t:dispatch g:dispatch
+    let [what, why] = dispatch#focus(a:count)
+    let g:dispatch = what
+    echo 'Set global focus to ' . what
   elseif empty(a:args)
     let [what, why] = dispatch#focus(a:count)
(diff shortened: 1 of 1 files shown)


autoload/dispatch.vim
```diff
@@ -706,10 +706,15 @@ function! dispatch#focus_command(bang, args, count) abort
   if has_key(opts, 'directory')
     let args = dispatch#dir_opt(opts.directory) . args
   endif
-  if empty(a:args) && a:bang
+  if empty(a:args) && a:bang && a:count < 0
     unlet! w:dispatch t:dispatch g:dispatch
     let [what, why] = dispatch#focus(a:count)
     echo 'Reverted default to ' . what
+  elseif empty(a:args) && a:bang && a:count > 0
+    unlet! w:dispatch t:dispatch g:dispatch
```
teoljungberg: This is how I deemed the behavior, but I'm happy to discuss it.

teoljungberg:
This is currently undocumented, and if we decide that this is something we want - I'll happily update the docs.

tpope:
I forgot about the global/window distinction when I made that suggestion. This means `:FocusDispatch! foo` will set a window local default but `:.FocusDispatch!` will set it globally. Unacceptably confusing.

I'm pushing a change to resolve `:Focus :.Dispatch` and a few other special cases. I believe that will adequately address your use case.

teoljungberg:
I see, your suggestion makes a lot of sense.
I do like the interface you made, it feels more intuitive. Thank you!

Is it possible to get tab-completion for commands entered after the`:Focus :`?

tpope:
Something could probably be stitched together using `getcompletion(..., 'command')`. Eagerly awaiting your PR.

teoljungberg:
Gladly!

After some tinkering with this, I ran into a solution that might work - I do need some vim-regex help though.

I've tinkered with this regex http://www.rubular.com/r/92WAfwu0Q4 to be able to extract the command (match-group 2) and pass that into `getcompletion(..., 'command')`. But I can't seem to get it to play nicely with vim's regex engine.

I've escaped the capture groups, but I've come up short.

Any ideas?

PR #231, opened 2018-03-16
Add `:Focus` completion for commands

teoljungberg:
This commit makes `:Focus :some-command` perform tab completion on the
matching commands, all while taking into consideration counts by line
numbers or relative counts such as `.`

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -509,6 +509,13 @@ function! dispatch#command_complete(A, L, P) abort
     let as = {'dir': 'directory'}
     let results = filter(['-compiler=', '-dir='],
           \ '!has_key(opts, get(as, v:val[1:-2], v:val[1:-2]))')
+  elseif a:A =~# '^:' && exists('*getcompletion')
+    let matches = matchlist(a:A, '^:\([.$]\|\d\+\)\=\(\a.*\)')
+    if len(matches)
+      let results = map(getcompletion(matches[2], 'command'), '":".matches[1]."".v:val')
+    else
+      let results = []
+    endif
   elseif a:A =~# '^\%(\w:\|\.\)\=[\/]'
     let results = s:file_complete(a:A)
   else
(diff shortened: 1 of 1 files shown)


teoljungberg:
As discussed in https://github.com/tpope/vim-dispatch/pull/230#issuecomment-373415280

tpope:
Darn, I thought this would get us results like `-complete=command` but it doesn't even handle additional arguments, what a bummer.

Try a pattern like `'^:\([.$]\|\d\+\)\=\(\a.*\)`. That'll handle `:Dispatch`, `:.Dispatch`, `:$Dispatch`, and `:9Dispatch`, the same 4 cases that I added support for.

Double check that you're handling an unsupported variants like `:.,+7Dispatch` (returning zero completion results is fine). Looks like your current construct falls back to `''` if the regex doesn't match, which means `getcompletion('', 'command')`, no bueno.

Lastly, add `&& exists('*getcompletion')` to the conditional so it doesn't blow up on older Vim versions.

teoljungberg:
All done!

> Double check that you're handling an unsupported variants like :.,+7Dispatch (returning zero completion results is fine).

Check, and check.

autoload/dispatch.vim
```diff
@@ -509,6 +509,13 @@ function! dispatch#command_complete(A, L, P) abort
     let as = {'dir': 'directory'}
     let results = filter(['-compiler=', '-dir='],
           \ '!has_key(opts, get(as, v:val[1:-2], v:val[1:-2]))')
+  elseif a:A =~# '^:' && exists('*getcompletion')
+    let matches = matchlist(a:A, '^:\([.$]\|\d\+\)\=\(\a.*\).')
```
tpope: You included the period from the end of the sentence!

autoload/dispatch.vim
```diff
@@ -509,6 +509,13 @@ function! dispatch#command_complete(A, L, P) abort
...
+    if len(matches[1:-1]) > 0
```
tpope: Just check `len(matches)`.

tpope: And stylistically I favor omitting the ` > 0`.

teoljungberg:
I've squashed and fixed your comment in one schwoop.

Thank you for the help and guidance!

tpope:
I removed a redundant `.""` and squashed it in.

PR #232, opened 2018-04-12
Fix tempname() results when we get short paths

filcab:
This is a problem on Windows, where the temporary file's directory is inside the user's homedir, which might require a long path component along the way.

Similar to https://github.com/tpope/vim-fugitive/pull/965

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -8,6 +8,14 @@ let g:autoloaded_dispatch = 1
 
 " Utility {{{1
 
+function! dispatch#tempname() abort
+  let temp = tempname()
+  if has('win32')
+    return fnamemodify(fnamemodify(temp, ':h'), ':p').fnamemodify(temp, ':t')
+  endif
+  return temp
+endfunction
+
 function! dispatch#uniq(list) abort
   let i = 0
   let seen = {}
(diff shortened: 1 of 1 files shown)


autoload/dispatch.vim
```diff
@@ -263,7 +271,7 @@ function! s:extract_opts(command) abort
       let val = 1
     endif
     if opt ==# 'dir' || opt ==# 'directory'
-      let opts.directory = fnamemodify(expand(val), ':p:s?[^:]\zs[\\/]$??')
+      let opts.directory = dispatch#fixtempname(fnamemodify(expand(val), ':p:s?[^:]\zs[\\/]$??'))
```
tpope: What's this one for?

filcab: It was from an earlier attempt at fixing the problem. That one doesn't seem to be needed, so I removed it.

tpope:
Let's name the function `dispatch#tempname()` and drop the argument.

filcab:
> Let's name the function dispatch#tempname() and drop the argument.
Done, thank you.

autoload/dispatch.vim
```diff
@@ -8,6 +8,15 @@ let g:autoloaded_dispatch = 1
 
 " Utility {{{1
 
+function! dispatch#tempname() abort
+  let l:temp = tempname()
```
tpope: Drop the `l:` prefixes to stylistically match the rest of the plugin.

autoload/dispatch.vim
```diff
@@ -8,6 +8,15 @@ let g:autoloaded_dispatch = 1
...
+  if has('win32')
+    " Fix shortpaths being returned on windows (e.g: if the user's name is long)
```
tpope: Move this comment in to the commit message.

autoload/dispatch.vim
filcab: Done

autoload/dispatch.vim
filcab: Done

PR #236, opened 2018-05-24
option to disable default mappings

flipcoder:


tpope:
Which one is irking you? I'll add an option before release but until then I want to hear complaints.

flipcoder:
No issue with the mapping choices, they just happen to clobber existing custom mappings that I was using previously.  Whenever that happens I look for a way to override it in the plugin.  The other plugins that I have to do this with have a similar naming and check for overriding defaults.

tpope:
I wanna hear about custom mappings this conflicts with because if there are common ones I'll avoid one.

flipcoder:
For me it was apostrophe (') which I have mapped as a fast scroll motion, so when the plugin defaults there is a delay.  I used my own custom macro setup before vim (it was very similar vim but it had operators on left hand and motions on the right hand, so the keys were different).  When i transitioned to vim I remapped those so they wouldn't make much sense for anyone else.

tpope:
You might appreciate `<nowait>`.

Issue #213, opened 2017-08-31
Add support for :terminal


gauteh:
Is it possible to use vims `:terminal` as adapter?

Related: #124

gauteh:
Uh, perhaps this is covered by #210 and #193.

tpope:
I'm thinking that the official interface is going to be `job_start()` for `:Dispatch`, and `:terminal` for `:Start`.

teoljungberg:
Any ideas on how that approach will fit in when one wants to run a test with a debugger using a job rather than a terminal?

-teo

tpope:
Not very well. I was thinking of adding an interface to force `:Dispatch` to run as a `:Start`.

teoljungberg:
Something like `:Start :Dispatch` would make sense.

teoljungberg:
I hacked together a proof of concept for the `terminal` start adapter.

It's available here: https://github.com/tpope/vim-dispatch/commit/140a3472309d1c49e9b429656b2e223acfc8ed12

I'd like to get `:Start` to re-use the same window - and I'd like pointers to get it to do so.

@tpope If you them want to run with this POC to combine the `job` branch with this - I'm totally OK with that.

tpope:
Implement `#activate` to get reuse.

tpope:
In a stroke of genius, I have realized `:{range}Spawn` has no sensible interpretation, and can be used to spawn the default dispatch.

teoljungberg:
I got `#activate` to play nicely with re-using already created windows: https://github.com/teoljungberg/vim-dispatch/commit/640c0397fb9fd411ac2a899c46aa4b6137194616
Any pointers on how to improve it would be greatly appreciated.

I'm going to use this setup of the [job](https://github.com/tpope/vim-dispatch/tree/job) branch with my branch of [terminal-as-start](https://github.com/tpope/vim-dispatch/compare/master...teoljungberg:terminal-as-start) to see if I catch any irregularities. But I'm excited to have this feature into vim-dispatch.

Have we gotten any further in testing the `job` branch on neovim to get closer to a finished feature?

tpope:
Posted an update to #193.

tpope:
Using the tab page number is too brittle. You'll need to grab the buffer number and then hunt down the proper tab/window upon switching.

teoljungberg:
[...]

PR #250, opened 2018-10-02
Add terminal handler for `:Start` actions

teoljungberg:
This PR adds support for using vim's `:terminal` for `:Start` and
`:Spawn` actions. 

Left to do:
- [ ] Support for NeoVim.


teoljungberg:
Since the [job] branch is not supported on NeoVim just yet, I thought I'd open this PR without support for NeoVim aswell. I've been using this without any real concerns for the past few weeks. 

[job]: https://github.com/tpope/vim-dispatch/tree/job

teoljungberg:
Should this PR be pointing towards the `job` branch or `master`? I'm up for either.

ejones:
Hey @teoljungberg - I've been using your POC from #213 locally for a couple months and hacking on it intermittently. I think you can just use [`'switchbuf'`](http://vimdoc.sourceforge.net/htmldoc/options.html#'switchbuf') to achieve the tab/window reuse in `#activate`. I've been using something like this locally and it seems to do the trick:
```vim
function! dispatch#terminal#activate(pid) abort
  " ...
      let swb = &l:switchbuf
      try
        let &l:switchbuf = 'useopen,usetab'
        silent exec 'tab sbuffer' buf_id
      finally
        let &l:switchbuf = swb
      endtry
  " ...
endfunction
```

Also, a small nit: `bufnr()` returns -1 when it can't find a buffer. So in your code where you do `elseif buf_id ... else ...`, I'm not sure that `else` branch will be triggered, since -1 is not falsy - unless I'm missing something.

One other thing I noticed: background invocation doesn't seem to work in this code. `:Start` and `:Start!` both end up with the new tab page focused. In `#handle` you have:
```vim
  silent exec 'tab sbuffer ' . buf_id
  if !a:request.background | wincmd p | endif
```
That seems to unconditionally drop the editor into a new tab, and `wincmd p` doesn't seem to have an effect since there's only one window in that tab. I think the `tab sbuffer` command should be conditional on `request.background`:
```vim
if !a:request.background
  silent exec 'tab sbuffer' buf_id
endif
```

teoljungberg:
Hi @ejones 

Thanks for the comments, it would have been easier to address them inline - but I'll do my best to do so here.

>  think you can just use 'switchbuf' to achieve the tab/window reuse in `#activate` ...

That's a neat trick, I'm into it as it! I updated to use it in e3b372a. I also fixed the `bufnr` handling in the same schwoop, thanks for catching it!

> One other thing I noticed: background invocation doesn't seem to work in this code. `:Start` and `:Start!` both end up with the new tab page focused.

The idea is to spawn a tab for `:Start` and focus on it, and `:Start!` should start a tab and not focus on it. Does that make sense?

But you are absolutely right about the behavior - it didn't work as expected. It seems like `:tabprevious` does the trick. I updated to use that in c8c6f3e.

Thanks for testing, and the feedback!

tpope:
@teoljungberg `job` gets rebased, I'd stick to `master` for now.

teoljungberg:
Is there anything you want me to revise to get this ready to be merged after the job branch has been merged? I saw in https://github.com/tpope/vim-dispatch/pull/193#issuecomment-493651760 that the job support is getting near completion.

I remember you commenting in a twitter thread that the use of the name for the tab would be fragile. Any feedback on having a better implementation in place would be greatly appreciated!

tpope:
Actually, that raises the question of what happens when you have a title collision. If you do the same `:Start somecommand` in two different directories, what happens?

teoljungberg:
Good point!

I `lcd`'d ito directory one and ran `:Start irb` and a new tab was opened. Opened a new tab, `lcd`'d into a new directory and ran the same command and it started a new tab with `irb` in it. I then switched to the first tab, and ran `:Start irb` again and it changed to the tab belonging to directory 2.

So you're absolutely right it's a collision. Any ideas on how we can resolve it? Saving the buffer number somehow and then switching to that?

teoljungberg:
I updated my implementation to save the id of the buffer in the request, but I get a feeling that it's tacking too much onto the request.

Let me know what you think @tpope.

tpope:
Look into using `s:postfix()`. That's how we disambiguate between duplicate quickfix lists.

teoljungberg:
I'm not sure what `s:postfix()` would give here, am I missing something?

https://github.com/teoljungberg/vim-dispatch/blob/c7e1ea1452964bee415690fa3ab749f3a26273b8/autoload/dispatch.vim#L354-L357

tpope:
Do a `:Dispatch` and look at the window title (`w:quickfix_title`). It's not just for show, it's used to decide if the current quickfix list corresponds to a given job. Perhaps something similar could be done with `term_name`?

teoljungberg:
Got it, I couldn't access the function from the current context so I copied it in. But `s:postfix()` returns the pid in it's output, or `?` when none exist. And when we set the title of the terminal - we don't have a pid to attach - so I couldn't get re-attach to work.

Is this preferred over saving the buffer id either on the request, or as a buffer variable in the window and find it by that?

tpope:
Oh dang, a real catch 22.

I'd say ditch `s:waiting` entirely, and iterate over `term_list()` looking for a `job_info(term_getjob(bufnr)).process` match. (This is basically your buffer variable idea, only without the need for a variable.) If that's not viable, you could structure `s:waiting` something like `{pid: [bufnr, request]}`, so at least the request object isn't tainted.

autoload/dispatch/terminal.vim
```diff
@@ -0,0 +1,71 @@
...
+    return 0
+  endif
+
+  call dispatch#autowrite()
+
+  let options = {
+        \ 'exit_cb': function('s:exit'),
+        \ 'hidden': 1,
+        \ 'term_name': a:request.title,
+        \ 'term_finish': (a:request.background ? 'open' : 'close'),
```
tpope: `background` isn't the concern here, it's `wait`. `always` and `never` correspond to `open` and `close`. I'm not sure what to do about `error`.

tpope: Oh, could the exit callback close the window if the exit status is 0?

teoljungberg: That sounds like a plan, but I got a weird behavior when I close the buffer in the exit callback in that it puts the focus on another window - and not the one I started out in.

teoljungberg:
I updated to use your suggested approach to find the buffer in 9274db2.

autoload/dispatch/terminal.vim
teoljungberg: Worth pointing out, this is the behavior even with the code in this PR. I'd like to iron that out before this is anywhere ready to get merged.

teoljungberg: A solution for that would be to tack on the buffer id of where the `:Start` or `:Spawn` came from to the request and the leverage the `switchbuf` trick we have here to navigate to it after closing the buffer. Ideas?

teoljungberg: Since the terminal will be opened in a new tab, and those are index-wise + 1 to the current tab, I can switch back to the old one based on that assumption.

I did so in 8d6a90d

teoljungberg: ...and I closed the terminal window when the exit status is 0 in the same commit.

tpope:
Rebase this on `master` and I can start testing it in earnest. Prioritize it just below the `job` strategy.

teoljungberg:
Rebased and ready for testing.

autoload/dispatch/terminal.vim
```diff
@@ -0,0 +1,80 @@
...
+function! s:exit(job, status) abort
+  let pid = job_info(a:job).process
+  let request = s:waiting[pid]
+  call writefile([a:status], request.file . '.complete')
+  if a:status == 0
+    let buf_id = s:buffer_for_pid(pid)
+    let terminal_window = tabpagewinnr(buf_id)
+    let current_window = terminal_window - 1
+    silent exec current_window . 'tabnext'
+    silent exec 'bdelete ' . buf_id
```
teoljungberg: This turned out to be trickier than I imagined, and I could use some help and guidance.

I did a verbose implementation of the functionality, and landed in this:

```diff
diff --git a/autoload/dispatch/terminal.vim b/autoload/dispatch/terminal.vim
index 2efc077..cf647cc 100644
--- a/autoload/dispatch/terminal.vim
+++ b/autoload/dispatch/terminal.vim
@@ -43,12 +43,19 @@ function! s:exit(job, status) abort
   let pid = job_info(a:job).process
   let request = s:waiting[pid]
   call writefile([a:status], request.file . '.complete')
-  if a:status == 0
-    let buf_id = s:buffer_for_pid(pid)
-    let terminal_window = tabpagewinnr(buf_id)
-    let current_window = terminal_window - 1
-    silent exec current_window . 'tabnext'
+  let previous_buffer_id = bufnr('#')
+  let buf_id = s:buffer_for_pid(pid)
+
+  if has_key(request, 'wait') && request.wait ==# 'always'
+  elseif has_key(request, 'wait') && request.wait ==# 'never'
+    silent exec 'bdelete ' . buf_id
+  elseif has_key(request, 'wait') && request.wait ==# 'error'
+    if a:status == 0
+      silent exec 'bdelete ' . buf_id
+    endif
+  elseif a:status == 0
     silent exec 'bdelete ' . buf_id
+    silent exec 'buffer ' . previous_buffer_id
   endif
   unlet! s:waiting[pid]
 endfunction
```

The cases that I imagined where:

- `:Start exit 0`, should start a new tab and close the tab when the command finishes.
- `:Start -wait=always exit 0`, should start a new tab and not close the tab when the command finishes.
- `:Start -wait=never exit 1` should start a new tab and close it when the command finishes.
- `:Start -wait=error exit 0` should start a new tab and close it when the command finishes.
- `:Start -wait=error exit 1` should start a new tab and not close it when the command finishes.

I got the `-wait` functionality in place, but the issue comes with when running a command that takes a few seconds (like running a test). As you mentioned @tpope we want to focus on the current buffer that is opened, but I couldn't get that to work. 

Do you have any suggestions?

tpope: When I change it to a simple `bdelete <bufnr>` it seems to be working as I expect. I assume what's throwing you is that if the terminal is focused when it exits, it switches to the next tab rather than the previous tab. This is dumb, but it's dumb in the standard Vim way, I think it's fine.

teoljungberg: Oh! That what was I was trying to resolve, to do the same as the tmux adapter.

If we want to focus back to the tab we came from, there's always adding the previous to the request and then leverage that.

But if we're fine with closing the tab and let the previous tab being focused - I'll make that work first

tpope: I think the way to solve that is for the terminal strategy to save `win_getid()` somewhere (ideally not on the request object), and then switch back to it on exit *only* if the terminal is currently the active window. Alternatively, we could just always force previous tab rather than next tab in that case, since 99% of the time that's what you want.

tpope:
Is there a better Vim built-in for switching out of the newly created terminal tab than `<C-W>:tabnext<CR>`?

autoload/dispatch/terminal.vim
tpope: To be clear, solving that is not a high priority, it's not a deal breaker for merging.

teoljungberg: I pushed a5df98e to work with the `-wait` flag, and simply closing the buffer with `:bdelete`.

autoload/dispatch/terminal.vim
```diff
@@ -0,0 +1,86 @@
...
+    if a:status == 0
+      silent exec 'bdelete ' . buf_id
+    endif
+  elseif a:status == 0
+    silent exec 'bdelete ' . buf_id
+  endif
+  unlet! s:waiting[pid]
+endfunction
+
+function s:buffer_for_pid(pid) abort
```
tpope: Add the usual `!`.

autoload/dispatch/terminal.vim
teoljungberg: I also did go with a verbose solution of walking through the different branches of the `-wait` flag. Let me know if you want me to tidy that up.

autoload/dispatch/terminal.vim
teoljungberg: Done in a5df98e.

teoljungberg:
> Is there a better Vim built-in for switching out of the newly created terminal tab than <C-W>:tabnext<CR>?

You mean rather than the `tab sbuffer <bufnr>` we're already doing?

tpope:
No I mean interactively. How the hell am I supposed to switch out of this terminal tab?

teoljungberg:
You mean that we should enter in the escape sequence for the user automatically or something similar? `<C-\><C-n>`

autoload/dispatch/terminal.vim
tpope: Not urgent but I'd like it tidied up before release. A conditional like `get(request, 'wait') is# 'never' || (get(request, 'wait') isnot# 'always' && a:status == 0)` ought to do it.

tpope:
This isn't a code question. Uninstall dispatch.vim entirely and replace `:Start` with `:tab terminal` and my question is the same: how the fuck do I switch out of the terminal?

teoljungberg:
Oh I see what you're saying. I thought `term_sendkeys` might be worthwhile looking into, but that seems to send the keys to the underlaying program - and not the terminal itself.

tpope:
This has nothing to do with `term_sendkeys`. I'm asking how to switch Vim tabs from inside a `:terminal`.

teoljungberg:
Besides your `<C-W>:tabprevious` idea, or my `<C-\><C-N>` to get into terminal-normal-mode - I don't know of a better way no.

teoljungberg:
And from Vim's help, that is what is suggested too:

> Use CTRL-W N (or 'termwinkey' N) to switch to Terminal-Normal mode.  Now the
contents of the terminal window is under control of Vim, the job output is
suspended.  CTRL-\ CTRL-N does the same.

tpope:
Looks like this was very recently addressed.

https://github.com/vim/vim/commit/72e83c1ae535e2ebc35b114d34d0a811eb62b068

autoload/dispatch/terminal.vim
```diff
@@ -0,0 +1,86 @@
...
+  if !has('terminal') || a:request.action !=# 'start'
+    return 0
+  endif
+
+  call dispatch#autowrite()
+
+  let options = {
+        \ 'exit_cb': function('s:exit'),
+        \ 'hidden': 1,
+        \ 'term_name': a:request.title,
```
tpope: This should be `a:request.expanded`. `a:request.title` is better suited for something short like a tmux window label.

In fact, let's make it `'!' . a:request.expanded` to match the usual convention.

autoload/dispatch/terminal.vim
```diff
@@ -0,0 +1,86 @@
...
+        \ 'term_finish': 'open',
+        \ }
+  let buf_id = term_start([&shell, &shellcmdflag, a:request.expanded], options)
```
tpope: If you make this `tab let ...`, you can avoid the whole hidden dance entirely.

autoload/dispatch/terminal.vim
```diff
@@ -0,0 +1,86 @@
...
+
+function! dispatch#terminal#handle(request) abort
+  if !get(g:, 'dispatch_experimental', 1)
+    return 0
+  endif
+  if !has('terminal') || a:request.action !=# 'start'
+    return 0
+  endif
+
+  call dispatch#autowrite()
```
tpope: Redundant.

autoload/dispatch/terminal.vim
teoljungberg: Done in b7e5609

autoload/dispatch/terminal.vim
teoljungberg: Really nice, done in 3f699a7.

autoload/dispatch/terminal.vim
teoljungberg: Killed in b8fb695

teoljungberg:
An idea I had - do we want to run the `:terminal` inside the same tab as the vanilla `:terminal`? And if you want to force it into a tab we can do `:tab Start` etc. That might simplify the closing-a-tab issues and UX we were discussing.

Or are we happy with forcing it into a tab by default?

tpope:
I had been considering that. Even if we don't change the default, we could also make `:leftabove Start` and the like do what I mean.

I don't see how it simplifies things though if we still have to support `:tab Start`.

tpope:
I think we should remove `term_name`. Having the shell in the name is ugly but I think the ugly truth is better than a lie. Plus the name doesn't get saved in the session, so dropping it will keep things consistent. Furthermore, we can use that consistent name in an `activate` workflow that works across reloads by hunting for the window rather than the pid.

I'd like to provide `:Start -noshell` or something to avoid the shell entirely, and then maybe even make that the default at some point. But that's a separate PR.

teoljungberg:
I added the `term_name` to give parity with the tmux adapter, and to make the windows more easily discoverable.

How would we use the default name of the terminal window in `activate`?

tpope:
Pushed my attempt to the `reuse-terminal` branch.

teoljungberg:
Hey @tpope - is this a PR we want to have move forward into a next version of dispatch? If so, is there something I can do to help bring it forward?

tpope:
Short answers are "yes" and "not yet". This has been creeping up on my priority list but I'm not quite ready to deal with it.

Well, I guess one thing you could do is see if you can get the code in the `reuse-terminal` branch to work with this PR. But it's not urgent.

tpope:
It's coming back to me. The biggest thing I _need_ is a Neovim version.

ranjithshegde:
> It's coming back to me. The biggest thing I _need_ is a Neovim version.

That would be a great thing to have. As of now `:Start` and `!` are not at all interactive and break the launching of some binaries in certain cases.

There is also this [PR](https://github.com/neovim/neovim/pull/10842) in the works, if that is of any help in developing this

tpope:
Next steps:

* Support the new `.mods` request property. Call `exe a:request.mods 'split'` before `term_start()`, and add `curwin` to `options`.
* Use `function('s:exit', [a:request])` to pass the request directly to the exit handler, rather than working backwards to it from the pid.
* Use `let a:request.bufnr = term_start(...)` to save the buffer number on the request, and use that when possible rather than `s:buffer_for_pid()`.

These changes will reduce the amount of Vim specific code and make it easier to support Neovim.

teoljungberg:
> Support the new .mods request property. Call exe a:request.mods 'split' before
> term_start(), and add curwin to options.

Done in f324538 and 12f7d4a

> Use function('s:exit', [a:request]) to pass the request directly to the exit
> handler, rather than working backwards to it from the pid.

Done in 56dbd76

> Use let a:request.bufnr = term_start(...) to save the buffer number on the
> request, and use that when possible rather than s:buffer_for_pid().

Done in 7793d0d, however, I feel that this block could be cleaned up - but I'm
unsure how to retrieve that information from the state of the world at this
point.

https://github.com/teoljungberg/vim-dispatch/blob/026b4efb3ca3dc7169bcbb103edf6063dc58f63a/autoload/dispatch/terminal.vim#L31-L35

autoload/dispatch/terminal.vim
```diff
@@ -0,0 +1,76 @@
...
+  if has_key(a:request, 'wait') && a:request.wait ==# 'always'
+  elseif has_key(a:request, 'wait') && a:request.wait ==# 'never'
+    silent exec 'bdelete ' . a:request.bufnr
+  elseif has_key(a:request, 'wait') && a:request.wait ==# 'error'
+    if a:status == 0
+      silent exec 'bdelete ' . a:request.bufnr
+    endif
+  elseif a:status == 0
+    silent exec 'bdelete ' . a:request.bufnr
+  endif
```
tpope, line 52: Add `let wait = get(a:request, 'wait', 'error')` above this and the rest of the code should collapse down to something much simpler.

Also, go ahead and try swapping in `bdelete!` here as it appears I will need that for Neovim.

tpope:
> Done in [7793d0d](https://github.com/tpope/vim-dispatch/commit/7793d0df11b9445d1c6e1b35a634f6517e27c5b8), however, I feel that this block could be cleaned up - but I'm unsure how to retrieve that information from the state of the world at this point.
> 
> https://github.com/teoljungberg/vim-dispatch/blob/026b4efb3ca3dc7169bcbb103edf6063dc58f63a/autoload/dispatch/terminal.vim#L31-L35

Leave it for now. It'll need to be restructured for Neovim anyways.

autoload/dispatch/terminal.vim
```diff
@@ -0,0 +1,76 @@
...
+  unlet! s:waiting[a:request.pid]
+endfunction
+
+function! dispatch#terminal#activate(pid) abort
+  if index(keys(s:waiting), a:pid) >= 0
```
tpope, line 57: `has_key(a:waiting, a:pid)`

autoload/dispatch/terminal.vim
```diff
@@ -0,0 +1,76 @@
...
+    let request = s:waiting[a:pid]
+
+    if request.bufnr
```
tpope, line 60: This can't be false.

autoload/dispatch/terminal.vim
tpope:
```vim
  let wait = get(a:request, 'wait', 'error')
  if wait ==# 'never' || (wait !=# 'always' && a:status == 0)
    silent exec 'bdelete! ' . a:request.bufnr
  endif
```

teoljungberg: You got it. 6ec244a

autoload/dispatch/terminal.vim
teoljungberg: Done 0229298

autoload/dispatch/terminal.vim
teoljungberg: b334331

autoload/dispatch/terminal.vim
```diff
@@ -0,0 +1,66 @@
...
+
+  let a:request.handler = 'terminal'
+
+  let options = {
+        \ 'exit_cb': function('s:exit', [a:request]),
+        \ 'term_name': '!' . a:request.expanded,
+        \ 'term_finish': 'open',
+        \ 'curwin': 1,
+        \ }
+  exe a:request.mods 'split'
```
teoljungberg, line 28: How should we handle `Start!` actions? Should they be opened in a tab that we focus away from, or should they be opened in a split we focus away from?

autoload/dispatch/terminal.vim
```diff
@@ -0,0 +1,66 @@
...
+  let a:request.bufnr = term_start([&shell, &shellcmdflag, a:request.expanded], options)
```
teoljungberg, line 29: I haven't found a good way if retrieving the bufnr and pid of a opened terminal in neovim that doesn't include finding the term windows and retrieving a buffer variable.

Do you have a good idea of how to achieve this @tpope ? If so, we can extract a `s:start` function that handles opening a terminal etc and returns the bufnr and pid for us to consume.

tpope: We get the job id back from `termopen()` which can be passed to `jobpid()`. For buffer number, there doesn't appear to be a general purpose mechanism, but the current buffer number immediately after `termopen()` should always be correct, I think.

Another caveat is we need to call `:new` rather than `:split` with Neovim because, well, it seems to take over the existing buffer? Sure, that's a totally reasonable behavior. :roll_eyes:

autoload/dispatch/terminal.vim
tpope: We should respect `request.mods`, I think, defaulting to a split we focus away from but supporting `tab` as well. Hopefully that doesn't complicate things too much.

autoload/dispatch/terminal.vim
tpope: This is what I hacked up locally:

```vim
  if has('nvim')
    exe a:request.mods 'new'
    let options = {
          \ 'on_exit': function('s:exit', [a:request]),
          \ }
    let job = termopen(a:request.expanded, options)
    let a:request.bufnr = bufnr('')
    let a:request.pid = jobpid(job)
    startinsert
  else
    exe a:request.mods 'split'
    let options = {
          \ 'exit_cb': function('s:exit', [a:request]),
          \ 'term_name': '!' . a:request.expanded,
          \ 'term_finish': 'open',
          \ 'curwin': 1,
          \ }
    let a:request.bufnr = term_start([&shell, &shellcmdflag, a:request.expanded], options)
    let job = term_getjob(a:request.bufnr)
    let a:request.pid = job_info(job).process
  endif

  let s:waiting[a:request.pid] = a:request
  call writefile([a:request.pid], a:request.file . '.pid')
```

I haven't beat on it too hard yet. You'll also need to add `, ...` to the `s:exit()` function signature.

autoload/dispatch/terminal.vim
teoljungberg: I'll see what I can work up.

autoload/dispatch/terminal.vim
teoljungberg: Aye that's similar to what I had besides a commented out `" let a:request.bufnr = ... ` section.

Another things that surprised me is that `has('terminal')` on neovim doesn't return >0. I supposed it is because neovim always has a terminal?

autoload/dispatch/terminal.vim
teoljungberg: Done in a43df4b

autoload/dispatch/terminal.vim
teoljungberg: I backported your suggested fix in b9fa6ed, and improved the terminal check for nvim and vim with terminal in 60e819e

autoload/dispatch/terminal.vim
```diff
@@ -0,0 +1,94 @@
...
+  if has('nvim')
+    exe a:request.mods 'new'
+    let options = {
+          \ 'on_exit': function('s:exit', [a:request]),
+          \ }
+    let job = termopen(a:request.expanded, options)
+    let a:request.bufnr = bufnr('')
+    let a:request.pid = jobpid(job)
+
+    if a:request.background
```
teoljungberg, line 32: This block is repeated for both nvim and normal vim, as I'm unsure if there are mechanism we want to use that exist in vim that doesn't exist in nvim and vice versa. For now, they are the same.

autoload/dispatch/terminal.vim
```diff
@@ -0,0 +1,94 @@
...
+          \ 'term_name': '!' . a:request.expanded,
+          \ 'term_finish': 'open',
+          \ 'curwin': 1,
+          \ }
+    let a:request.bufnr = term_start([&shell, &shellcmdflag, a:request.expanded], options)
+    let job = term_getjob(a:request.bufnr)
+    let a:request.pid = job_info(job).process
+
+    if a:request.background
+      if a:request.mods ==# 'tab'
```
tpope, line 54: Use `=~# '\<tab\>'` to catch more esoteric uses like `:silent tab Start`.

autoload/dispatch/terminal.vim
```diff
@@ -0,0 +1,94 @@
...
+        tabprevious
+      else
+        wincmd w
+      endif
```
tpope, line 58: Need to double check versioning but `win_getid()` plus `win_gotoid()` should be able to replace both halves of this conditional far more robustly.

teoljungberg: As far as I managed to dig, `win_getid()` been in vim 7.4.1557:

https://github.com/vim/vim/commit/86edef664efccbfe685906c854b9cdd04e56f2d5

`win_gotoid()` was included in the same version.

Is that recent enough that we can use it?

autoload/dispatch/terminal.vim
teoljungberg: Implemented as 5f6d228, but is made redundant after df369bb

autoload/dispatch/terminal.vim
teoljungberg: Done in df369bb, if the feature is too recent we'll investigate a plan-b.

tpope: That's older than `:terminal` so yes, we can definitely use it.

autoload/dispatch/terminal.vim
```diff
@@ -0,0 +1,88 @@
...
+if !exists('s:waiting')
+  let s:waiting = {}
+endif
+
+function! dispatch#terminal#handle(request) abort
+  if !get(g:, 'dispatch_experimental', 1)
+    return 0
+  endif
+  let has_terminal = (has('terminal') || has('nvim'))
+  if !has_terminal || a:request.action !=# 'start'
```
tpope, line 17: Every time I see the `has_terminal` variable I ask myself "is this variable used elsewhere?" So let's make it one line.

autoload/dispatch/terminal.vim
```diff
@@ -0,0 +1,88 @@
...
+          \ 'term_finish': 'open',
+          \ 'curwin': 1,
+          \ }
+    let a:request.bufnr = term_start([&shell, &shellcmdflag, a:request.expanded], options)
+    let job = term_getjob(a:request.bufnr)
+    let a:request.pid = job_info(job).process
+
+    if a:request.background
+      call win_gotoid(winid)
+    endif
```
tpope, line 53: Let's pull this out of the conditional, along with the corresponding assignment.

tpope:
I think this is just about ready. Add this snippet to the docs above "Windows" (same order as `g:dispatch_handlers`):

```
Terminal ~

Uses the respective |:terminal| features of Vim and Neovim.
```

autoload/dispatch/terminal.vim
teoljungberg: Sure, done in b946ce0

autoload/dispatch/terminal.vim
teoljungberg: Sure thing, I kept a `if !a:request.background | startinsert | endif` inside the neovim block so we'd only go into insert mode for the terminal buffer when it's backgrounded and not the current window

teoljungberg: 294dded

teoljungberg:
Have you noticed any rumblings around this @tpope ? Or should we start ripping out / disabling the low priority adapters?

tpope:
Not in a rush, but I guess we can go ahead with the first removal.

Issue #179, opened 2016-07-18
Unable to use user installed shell


smutch:
I use my own build of zsh on a system where I do not have root access.  This shell further depends on my own build of pcre.  When the shell is invoked by dispatch, a clean environment is is used and therefore my shell is unable to launch (due to not having access to pcre).

I've overcome this by adding a config variable `g:dispatch_extra_env_vars` (see https://github.com/smutch/vim-dispatch/commit/fc01b5e2955843353aba45f27e70a6e69ea26d59), however, I wanted to open an issue to ensure that there is no better way to do this and to see if this would be considered useful enough for a PR.

tpope:
Could you handle this on your end using a wrapper script?

smutch:
Sure.  I'd definitely be happy to try that as an alternative (I know this is probably a bit of an edge case).  

I'm not really sure how to do this though.  Would you have any high-level recommendations of how to proceed?  How would I write a wrapper to achieve this and use it?

tpope:
I think you want something like

```sh
#!/bin/sh
MYVAR1=a MYVAR2=b exec myshell
```

smutch:
Works like a charm.  Thanks for that!

tpope:
I accidentally left out `"$@"` at the end of the last line.

PR #252, opened 2018-11-24
Keep the USER environment variable in isolated command

LnL7:
This prevents issues similar to #179 if sections in bashrc/profile
of the default-shell in tmux assume these are available.

There might be a better way around this, changing the tmux configuration to start a non longin shell avoids the issue but that's undesirable for other panes. Not starting a login shell for the dispatch panes would also solve the particular case I ran into.

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -318,7 +318,7 @@ function! dispatch#set_title(request) abort
 endfunction
 
 function! dispatch#isolate(request, keep, ...) abort
-  let keep = ['SHELL'] + a:keep
+  let keep = ['SHELL', 'USER'] + a:keep
   let command = ['cd ' . shellescape(getcwd())]
   for line in split(system('env'), "\n")
     let var = matchstr(line, '^\w\+\ze=')


tpope:
`$HOME` appears all over my login files and I've never had an issue with either zsh or bash. What am I missing?

LnL7:
Oh indeed, it's only USER. I assumed both where gone and added it since it's also pretty common. Don't really understand why this happens tho.

tpope:
So `USER` is nonstandard, does it work with `LOGNAME`?

tpope:
Also tmux doesn't use a login shell for `tmux split-window 'some command'`. I think you've got something else going on.

LnL7:
I see what's going on now, I have some variables in zshenv that depend on USER. These are used later on resulting in weird warnings. Looks like LOGNAME is always available, but only for zsh so that's not a generic solution.

tpope:
`LOGNAME` is an old UNIX thing that's available everywhere AFAIK. Also most shells don't have a universally loaded file like `.zshenv` to hang yourself with.

PR #267, opened 2019-03-26
Make expandable substitution string case sensitive

### LINKED ISSUE #266: Dispatch doesn't like text after %:S in makeprg
I am using pandoc to compile and open Markdown files, like this:

```vim
CompilerSet makeprg=pandoc\ -s\ --metadata\ pagetitle=\"%:t:r\"\ -o\ %:t:r.html\ %:S\ &&\ open\ %:t:r.html
```

However, when I run `:Make` on a Markdown file, the command that is actually run is
```vim
:Dispatch pandoc -s --metadata pagetitle="README" -o README.html 'README.md'README.html
```
I worked through the source of vim-dispatch and I tracked down the problem to the `dispatch#expand` function. The substitute call was matching `%:S && open` as a submatch and sending that to the `s:expand` function, so `%:S && open ` was being replaced with `README.md`.

This is due to part of the `s:expandable` pattern; particularly, this part (in the `s:flags` portion):
```vim
:g\=s\(.\).\{-\}\1.\{-\}\1
```

This is matching on `:S open ` when it shouldn't be. It turns out, the solution is as simple as adding `\C` to the beginning of the `s:expandable` string:
```vim
let s:expandable = '\C\\*\%(`[+-]\==[^`]*`\|' . s:var . s:flags . '\)'
```
which makes the expansion case sensitive.

I created a PR to fix this in #267.

gpanders:
Closes #266

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -88,7 +88,7 @@ function! s:expand(expr, dispatch_opts) abort
 endfunction
 
 let s:flags = '<\=\%(:[p8~.htre]\|:g\=s\(.\).\{-\}\1.\{-\}\1\)*\%(:S\)\='
-let s:expandable = '\\*\%(`[+-]\==[^`]*`\|' . s:var . s:flags . '\)'
+let s:expandable = '\C\\*\%(`[+-]\==[^`]*`\|' . s:var . s:flags . '\)'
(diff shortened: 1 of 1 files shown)


tpope:
Good catch, thanks!

PR #273, opened 2019-06-09
Mention default maps also in README

kiryph:


README.markdown
```diff
@@ -97,7 +97,14 @@ can set it interactively, or in an autocommand:
...
+`:Dispatch` makes a great map. By default dispatch.vim already provides
+amongst others
+
+- `` `<CR>                   :Dispatch<CR>``
+- `` `<Space>                :Dispatch<Space>``
+
+All default mappings can be found under `:h dispatch-maps`. However, you
+can of course pick your own one, e.g.
 
     nnoremap <F9> :Dispatch<CR>
```
tpope: Let's ditch this example entirely. Anyone who has defined a map before should be able figure this out pretty easily.

README.markdown
```diff
@@ -97,7 +97,14 @@ can set it interactively, or in an autocommand:
 
 If no `b:dispatch` is found, it falls back to `:Make`.
 
-`:Dispatch` makes a great map:
+`:Dispatch` makes a great map. By default dispatch.vim already provides
+amongst others
+
+- `` `<CR>                   :Dispatch<CR>``
+- `` `<Space>                :Dispatch<Space>``
```
tpope: Let's limit this to `` `<CR> `` only, and combine it all into one paragraph so it flows better.

Alternatively, I was thinking of breaking maps out into their own section and then we can go into more detail. But that raises more issues like the fact we don't explain what `:Spawn` is, so we should probably save that for another day.

README.markdown
```diff
@@ -97,7 +97,14 @@ can set it interactively, or in an autocommand:
 
 If no `b:dispatch` is found, it falls back to `:Make`.
 
-`:Dispatch` makes a great map:
+`:Dispatch` makes a great map. By default dispatch.vim already provides
```
tpope: A minor nit I know, but since you're making changes anyways, I double space after periods in Markdown for Vim related reasons. http://stevelosh.com/blog/2012/10/why-i-two-space/ (tl;dr double spacing lets Vim handle `Mr. Smith` correctly for "sentence" operations like `das`.)

tpope:
I noticed this was in need of an update last night while cutting a release. Thanks for stepping up!

kiryph:
I force-pushed your requested changes. I also removed the list environment.

I think a separate section for the great maps of dispatch would be nice. More favourites of mine are `m<cr>` and `m<space>` (even though I setup `Make` on save in more and more situations).

> Thanks for stepping up!

I am always glad to see updates for my favourite plugins. I am actually waiting for some time to see the job PR merged to master.

Thank you very much for your long-term support for your plugins.

README.markdown
kiryph: Done

README.markdown
kiryph: Done and unneccesary list removed. (separate section is left for another day)

README.markdown
kiryph: Done

tpope: I mean one paragraph as in something like

    provides `` `<CR>`` as a map for `:Dispatch<CR>` among others

That grammar is a bit weird so feel free to tweak it further.

kiryph: I changed it to a single paragraph consisting of three sentences:

> `:Dispatch` makes a great map.  By default dispatch.vim provides `` `<CR>`` for
`:Dispatch<CR>`.  You can find all default maps under `:h dispatch-maps`.

I tried to make the grammar less weird. However, I am not a native speaker. Therefore, I am not really confident in making a decision about what sounds weird and what not.

tpope:
The `m` maps are some of my favorites too, probably because they're so easy to type!

kiryph:
>  probably because they're so easy to type!

I agree on this very much. This is a good reason to advertise them in the README with a dedicated section. Finding good mappings is for beginners or actually every vim user difficult.

IMHO they are so good that they could be used by people who do not use vim-dispatch and just run `:make`.

 Advocating good choices helps to establish good practices within the vim community. Similar how vim-commentary and tcomment share today `gc`.

PR #278, opened 2019-07-12
Add xprop fallback for windowid

### LINKED ISSUE #277: X11 skipped on neovim tui
Due to https://github.com/neovim/neovim/issues/3626

Workaround: Look at $WINDOWID as well

SevereOverfl0w:
Some popular terminals, e.g. gnome-terminal do not set $WINDOWID.
Additionally, the Neovim TUI does not yet support $WINDOWID.

Fix #277

### DIFF
--- a/autoload/dispatch/x11.vim
+++ b/autoload/dispatch/x11.vim
@@ -5,12 +5,19 @@ if exists('g:autoloaded_dispatch_x11')
 endif
 let g:autoloaded_dispatch_x11 = 1
 
+function! s:windowid()
+  if executable('xprop')
+    return matchstr(system("xprop -root _NET_ACTIVE_WINDOW"), '0x\x\+')
+  endif
+endf
+
 function! dispatch#x11#handle(request) abort
   if $DISPLAY !~# '^:'
     return 0
   endif
+  let windowid = s:windowid()
   if get(a:request, 'background') &&
-        \ (!v:windowid || !executable('wmctrl'))
+        \ (empty(windowid) || !executable('wmctrl'))
     return 0
   endif
   if exists('g:dispatch_terminal_exec')
@@ -23,16 +30,17 @@ function! dispatch#x11#handle(request) abort
     return 0
   endif
   if a:request.action ==# 'start'
-    return dispatch#x11#spawn(terminal, dispatch#prepare_start(a:request), a:request)
+    return dispatch#x11#spawn(terminal, dispatch#prepare_start(a:request), a:request, windowid)
   else
     return 0
   endif
 endfunction
 
-function! dispatch#x11#spawn(terminal, command, request) abort
+function! dispatch#x11#spawn(terminal, command, request, windowid) abort
   let command = dispatch#set_title(a:request) . '; ' . a:command
   if a:request.background
-    let command = 'wmctrl -i -a '.v:windowid . ';' . command
+    let command = 'wmctrl -i -a '. a:windowid . ';' . command
+    echom command
   endif
   call system(a:terminal . ' ' . dispatch#shellescape(&shell, &shellcmdflag, command). ' &')
   return 1


autoload/dispatch/x11.vim
```diff
@@ -5,12 +5,24 @@ if exists('g:autoloaded_dispatch_x11')
...
+  endif
+endf
+
 function! dispatch#x11#handle(request) abort
   if $DISPLAY !~# '^:'
     return 0
   endif
   if get(a:request, 'background') &&
-        \ (!v:windowid || !executable('wmctrl'))
+        \ (!s:windowid_available() || !executable('wmctrl'))
```
tpope: An `executable()` check is insufficient to tell if we have an active X display, etc. I think the check should just be something like:

```viml
let windowid = s:windowid()
if ...(!windowid ...)...
```

autoload/dispatch/x11.vim
```diff
@@ -5,12 +5,24 @@ if exists('g:autoloaded_dispatch_x11')
 endif
 let g:autoloaded_dispatch_x11 = 1
 
+function! s:windowid_available()
+  return v:windowid || executable('xprop')
+endf
+
+function! s:windowid()
+  if v:windowid
```
tpope: Let's kill `v:windowid` and go all in on `xprop`.

autoload/dispatch/x11.vim
```diff
@@ -5,12 +5,24 @@ if exists('g:autoloaded_dispatch_x11')
...
+    return v:windowid
+  elseif executable('xprop')
+    return substitute(system("xprop -root '\t$0' _NET_ACTIVE_WINDOW"), '.*\t', '', '')
```
tpope: This will include the trailing newline, which is probably fine until some dumb shell like `csh` or `fish` chokes on it. Try `matchstr(..., '0x\x\+')`.

Single quotes are also tempting shell fate. That `matchstr()` should make the `'\t$0'` unnecessary.

autoload/dispatch/x11.vim
SevereOverfl0w: There's already a display check done above.

autoload/dispatch/x11.vim
SevereOverfl0w: Using the '\t$0' removes the trailing newline.

Not familiar with any single quote issues, a quick tcsh/fish test seemed fine.  I think single quotes as ultimately escaped goes way back.

tpope: I'd still feel better with `shellescape()`. Who knows, maybe some poor fool will want to get this to work with `cmd.exe`.

autoload/dispatch/x11.vim
tpope: This doesn't check that the display is still running and we have permission to access it. Edge cases sure, but it's easy to guard against.

autoload/dispatch/x11.vim
SevereOverfl0w: It's gone anyway :)

autoload/dispatch/x11.vim
SevereOverfl0w: Because `!'foo'` is 1, this check will always pass.  Is there some trick to this, or is `== ''` fine? Pushing that suggestion, but not sure on best practice here.

SevereOverfl0w:
I've pushed up with changes requested. I've also passed through the windowid because it felt strange that we might potentially get `''` later, and didn't handle that.

autoload/dispatch/x11.vim
tpope: `!'0x1'` does fail, but an empty string check is definitely more appropriate. Use `!empty(windowid)` so the linter doesn't complain about `==` not being `==#`.

tpope:
Yes, passing through the windowid is appropriate, it also prevents an extra `system()` call which can be slow.

Issue #217, opened 2017-10-12
Empty output when using Fish shell


kyleholzinger:
I'm not sure if it's something that I'm doing or some config, but whenever I run any command with `:Dispatch`, the split is empty and doesn't show any output

tpope:
Try in 2 or more adapters (tmux, screen, plain terminal, etc) and see if it's specific to one of them.

kyleholzinger:
Ahh interesting, looks like it was an issue with the fact that I use fish. When I changed my default shell to bash it works @tpope

dduan:
I can confirm this.

AndreVolpi:
@tpope is there a fix to be able to use with fish and tmux?

kyleholzinger:
@AndreVolpi I set my default shell for all of vim to bash and that fixed it! Try putting this in your .vimrc:

```vimscript
set shell=bash
```

AndreVolpi:
Doing that makes dispatch open a new tmux window, which I understand was not the expected behavior.

kyleholzinger:
Doing what? setting the shell to bash? @AndreVolpi

AndreVolpi:
Yes, when I set Vim to use bash instead of fish, and I call dispatch, it opens a new window, instead of a new pane.

jsatk:
This seems related to my issue here: 
https://github.com/tpope/vim-dispatch/issues/243

PR #281, opened 2019-07-25
fix blank output in tmux when using fish shell

anark:
Hello,
When using dispatch with Fish shell set I get an empty output in all tmux windows while the command is running.

This fixes the output so the output shows while it is running.

### DIFF
--- a/autoload/dispatch/tmux.vim
+++ b/autoload/dispatch/tmux.vim
@@ -39,7 +39,7 @@ function! dispatch#tmux#handle(request) abort
 endfunction
 
 function! dispatch#tmux#make(request) abort
-  let pipepane = (&shellpipe ==# '2>&1| tee' || &shellpipe ==# '|& tee') && &shell !~# 'fish'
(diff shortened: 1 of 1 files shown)


tpope:
It was definitely added for a reason. Please blame the line and find the commit that added it so I can see if it's still relevant.

anark:
It was added in the commit to support fish shell(https://github.com/tpope/vim-dispatch/commit/4566b8892715361f9952fa4c29c05035fdede09d) and close the blank output issue(https://github.com/tpope/vim-dispatch/issues/217).  However when I run Dispatch without this change the output in the tmux pane is blank.  When I run it with this change the output displays correctly.

tpope:
Okay, since it was added with the rest of the Fish support I assume it was just based on my personal testing. I'm okay to remove it.

The real issue, though, appears to be a broken Fish behavior. The following command won't output anything until 10 seconds later after the entire `begin` block has finished running:

    begin; echo test; sleep 10; end | cat

Appears to be a longstanding issue. https://github.com/fish-shell/fish-shell/issues/1396

PR #285, opened 2019-10-22
Use v:echospace to avoid hit-Enter prompts

blueyed:
Fixes https://github.com/tpope/vim-dispatch/issues/166
Followup-to https://github.com/tpope/vim-dispatch/pull/167

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -370,7 +370,7 @@ function! s:echo_without_hitenter(msg, suffix) abort
     else
       let msg = a:msg
     endif
-    echom msg.' '.a:suffix
+    echo msg.' '.a:suffix
     return
   endif
 
@@ -408,7 +408,7 @@ function! s:echo_without_hitenter(msg, suffix) abort
   else
     let msg = a:msg
   endif
-  echom msg.' '.a:suffix
+  echo msg.' '.a:suffix
 endfunction
 
 function! s:dispatch(request) abort
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -363,7 +363,9 @@ function! s:postfix(request) abort
   return '(' . a:request.handler.'/'.(!empty(pid) ? pid : '?') . ')'
 endfunction
 
-function! s:echo_without_hitenter(msg, suffix) abort
+function! s:echo_truncated(msg, suffix) abort
+  redraw
+
   if exists('v:echospace')
     if strwidth(a:msg.' '.a:suffix) > v:echospace
       let msg = printf('%.*S...', v:echospace - 3 - len(a:suffix), a:msg)
@@ -374,7 +376,6 @@ function! s:echo_without_hitenter(msg, suffix) abort
     return
   endif
 
-  redraw
   let suffix_len = len(substitute(a:suffix, '.', '.', 'g'))
   let max_cmd_len = (&cmdheight * &columns) - 2 - suffix_len - 2
 
@@ -422,7 +423,7 @@ function! s:dispatch(request) abort
       let a:request.handler = handler
 
       " Display command, avoiding hit-enter prompt.
-      call s:echo_without_hitenter(':!'.a:request.expanded, s:postfix(a:request))
+      call s:echo_truncated(':!'.a:request.expanded, s:postfix(a:request))
       return response
     endif
   endfor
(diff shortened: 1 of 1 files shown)


autoload/dispatch.vim
```diff
@@ -363,6 +363,54 @@ function! s:postfix(request) abort
...
+    if &showcmd
+      let max_cmd_len -= 10
+      if !&ruler || last_has_status
+        let max_cmd_len -= 1
+      endif
+    endif
+  endif
+
+  let msg_len = len(substitute(a:msg, '.', '.', 'g'))
+  if msg_len > max_cmd_len
```
blueyed: Could also use `strwidth` maybe if available?

autoload/dispatch.vim
```diff
@@ -363,6 +363,54 @@ function! s:postfix(request) abort
...
 
+function! s:echo_without_hitenter(msg, suffix) abort
+  if exists('v:echospace')
+    if strwidth(a:msg.' '.a:suffix) > v:echospace
+      let msg = printf('%.*S...', v:echospace - 3 - len(a:suffix), a:msg)
+    else
+      let msg = a:msg
+    endif
+    echom msg.' '.a:suffix
+    return
```
blueyed: Could `echo`/`echom` directly (also below).

autoload/dispatch.vim
```diff
@@ -363,6 +363,54 @@ function! s:postfix(request) abort
   return '(' . a:request.handler.'/'.(!empty(pid) ? pid : '?') . ')'
 endfunction
 
+function! s:echo_without_hitenter(msg, suffix) abort
```
tpope: The message is "press enter", not "hit enter", but either way it's too much of a mouthful. Try `echo_truncated()`

autoload/dispatch.vim
```diff
@@ -363,6 +363,54 @@ function! s:postfix(request) abort
...
+    if strwidth(a:msg.' '.a:suffix) > v:echospace
+      let msg = printf('%.*S...', v:echospace - 3 - len(a:suffix), a:msg)
+    else
+      let msg = a:msg
+    endif
+    echom msg.' '.a:suffix
+    return
+  endif
+
+  redraw
```
tpope: Why does the old logic require a redraw but the new logic doesn't?

autoload/dispatch.vim
tpope: If it ain't broke don't fix it.

autoload/dispatch.vim
blueyed: Should be done with both probably.

autoload/dispatch.vim
```diff
@@ -363,6 +363,55 @@ function! s:postfix(request) abort
   return '(' . a:request.handler.'/'.(!empty(pid) ? pid : '?') . ')'
 endfunction
 
+function! s:echo_truncated(msg, suffix) abort
+  redraw
+
+  if exists('v:echospace')
+    if strwidth(a:msg.' '.a:suffix) > v:echospace
+      let msg = printf('%.*S...', v:echospace - 3 - len(a:suffix), a:msg)
```
tpope: This should use the `<` notation from the other format, not an ellipsis.

petobens:
Any chance of moving this forward @blueyed @tpope ? It would be a really nice addition... Thanks!

autoload/dispatch.vim
axelf4: Why is the `substitute()` needed here?

blueyed:
@petobens 
I feel like if I'd fix/amend it @tpope would do it himself anyway - so it's waiting on him.

petobens:
OK. Let's wait for @tpope then. Thanks

tpope:
If I was going to do it myself I would have done it myself rather than explicitly requesting changes.

autoload/dispatch.vim
tpope: Character length rather than byte length.

petobens:
Ok @blueyed can you wrap this up then? It's really annoying to press enter. Thank you so much in advance

telemachus:
@blueyed I merged this PR locally to see if it fixed the problem, and at least here it actually made it worse. I.e., with the earlier version I had the `Press ENTER` problem with some very long-named files, but now I have it with all files.

Perhaps it has something to do with my settings though? (`shortmess=atIF` and `cmdheight=1`, in case that helps.)

lourenci:
Hi, @blueyed. Is there some advance on this? Thank you for this.

PR #286, opened 2019-12-06
Reduce install instructions

casr:
Fairly pointless change so feel free to close ;)

Edit: probably should mention that the `git clone` makes the intermediate directories so that’s why you do not need to use `mkdir`


tpope:
Maybe someday. tpope/vim-unimpaired#179

PR #306, opened 2020-10-15
Add customizable message colors

Kamilcuk:
Allow for customing the message color of "Success/Failure/Aborted/Complete" echo msg below. By default colors are normal, so there should be no difference as to how it work currently. With that one can use:

```
highlight DispatchAbortedMsg   ctermbg=Red
highlight DispatchFailureMsg   ctermbg=Red
highlight DispatchSuccessMsg   ctermbg=Green
highlight DispatchCompleteMsg  ctermbg=Green
```

And I have super green message if something succeeded, and super red if it failed.

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -1194,15 +1194,6 @@ function! dispatch#complete(file, ...) abort
       let status = -1
       call writefile([-1], request.file . '.complete')
     endtry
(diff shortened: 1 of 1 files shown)


tpope:
I think I'm okay with this.

* I don't think we need to declare defaults.
* Use `echohl NONE`, not `Normal`, to turn off the highlighting.

PR #320, opened 2021-11-20
Allow custom message handler

p00f:
This allows the user to do what they want with the echo message

(my usecase is sending it to neovim's vim.notify which will use https://gitlab.com/HiPhish/desktop-notify.nvim to send desktop notifications)


p00f:
closed in favout of #321

PR #321, opened 2021-12-01
allow custom message handler

p00f:
This is #320 but from a separate branch

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -405,7 +405,13 @@ function! s:echo_truncated(left, right) abort
   if len(substitute(msg, '.', '.', 'g')) > max_len
     let msg = a:left . '<' . matchstr(a:right, '\v.{'.(max_len - len(substitute(a:left, '.', '.', 'g')) - 1).'}$')
   endif
-  echo msg
+
+  if exists('g:dispatch_message_handler')
+    call g:dispatch_message_handler(msg)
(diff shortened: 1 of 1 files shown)


p00f:
Is there interest in this/should I close it?

tpope:
I would rather see this done with a `QuickFixCmdPost` autocommand if at all possible. I haven't investigated what it would take to achieve this.

p00f:
I don't understand the `doautocmd` in autoload/dispatch.vim, the help for `doautocmd` says

`:do[autocmd] [<nomodeline>] [group] {event} [fname]` which means `dispatch-make-complete` is a file but `dispatch-make-complete` is only used in the `silent doautocmd Event dispatch-make-complete` position

p00f:
Closed by mistake

tpope:
`[fname]` isn't always a file.

p00f:
Do you mean calling `s:doautocmd('QuickFixCmdPost dispatch-echo')` from `s:echo_truncated` and adding the call to the handler in https://github.com/tpope/vim-dispatch/blob/527f200fb86c18d1cc7dc702d13f8818dced04ac/plugin/dispatch.vim#L100 ?

tpope:
More or less yes, but I think it should be higher level than that, maybe something like `dispatch-make-done`, with both `Pre` and `Post` versions.

PR #324, opened 2021-12-30
Check compiler_options.format instead of program

wodesuck:
Some compilers (such as gcc) don't set makeprg and comipler_options.program will not be set, that breaks :Dispatch when use those comiplers.
Check compiler_options.format instead fix it and also give an error on :Dispatch -compiler=invalid.


tpope:
This fixes one problem and causes another on `:Dispatch -compiler=gcc -- %` at https://github.com/tpope/vim-dispatch/blob/a58eaaa41f895fb729a36740a91608ed47706636/autoload/dispatch.vim#L881. I would propose changing that use of `request.program` to `get(request, 'program', get(request, 'compiler', '--))`. (That `--` should never be reached in practice, but avoids the error that `get(request, 'program', request.compiler)` would give us when `.compiler` is missing.)

wodesuck:
@tpope Fixed, pls review again.

PR #327, opened 2022-01-13
fix: do not attempt to close last window if quickfix with no errors

### LINKED ISSUE #323: Cannot close last window
When I have only the quickfix window open and run `:Make`, I get 
```
Error detected while processing function dispatch#compile_command:
line   91:
E444: Cannot close last window
```
caused by [this line](https://github.com/tpope/vim-dispatch/blob/3505862b3898be5db3c78ba1b92c703349478d68/autoload/dispatch.vim#L915). Is my workflow just weird or should this indiscriminate `cclose` be surrounded by a check if this is the only open window?

kaihowl:
Fixes #323

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -1362,7 +1362,11 @@ function! s:cwindow(request, all, copen, mods, event) abort
   if mods !~# 'aboveleft\|belowright\|leftabove\|rightbelow\|topleft\|botright'
     let mods = 'botright ' . mods
   endif
-  execute (mods) (a:copen ? 'copen' : 'cwindow') height
+  if a:copen
+    execute (mods) 'copen' height
+  elseif !was_qf || winnr('$') > 1
+    execute (mods) 'cwindow' height
+  endif
   if !was_qf && s:is_quickfix() && a:copen !=# -2
     wincmd p
   endif


autoload/dispatch.vim
```diff
@@ -1362,7 +1362,11 @@ function! s:cwindow(request, all, copen, mods, event) abort
   if mods !~# 'aboveleft\|belowright\|leftabove\|rightbelow\|topleft\|botright'
     let mods = 'botright ' . mods
   endif
-  execute (mods) (a:copen ? 'copen' : 'cwindow') height
+  if a:copen
+    execute (mods) 'copen' height
+  elseif !was_qf || winnr('$') > 1
+    execute (mods) 'cwindow' height
```
tpope, line 1368: "If we're not currently in the quickfix window, and there is only one window open, call `:cwindow`". Or to put it another way, "if the only window open isn't the quickfix window, don't open the quickfix window." This can't be right.

kaihowl: I believe the logic is sound. I also verified manually that this works as expected. If the only open window is indeed not the quickfix window, then `!was_qf` will be `true`. Therefore the or'ed clause will be true. Hence, we attempt to open the quickfix window.

The intention is to prohibit using `cwindow` when the quickfix window is the only open window. This is `!(was_qf && winnr('$')==1)`. Which is logically identical to `!was_qf || !winnr('$')==1`. And that can be simplified to `!was_qf || winnr('$')>1`. The last transformation assumes that `winnr` can never be `0`, which I believe is a sane assumption.

Stylistically, I can agree that the logic might be hard to grasp in this form. Should I rather use the original `!(was_qf && winnr('$')==1)`?

tpope: Dang this is hard to reason about. Your alternative form is a bit clearer, but I think I tend to re-arrange `!(a && b)` to `!a || !b` in my head anyways. I guess we'll go with this version.

PR #328, opened 2022-02-02
allow custom message handler

p00f:
#321 but I github wouldn't "continue" the PR from a deleted-then-created-again branch


p00f:
@tpope

p00f:
How will the user do what they want with the message then?

tpope:
Calling `dispatch#request()` is a good start.

p00f:
Thanks 👍🏽

PR #331, opened 2022-06-29
Let user to set copen split direction  (eg. vertical)

maho:
If you set eg `let g:quickfix_dispatch_split="vert to"` then `Copen` will open vertical panel on the left.

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -1361,7 +1361,6 @@ function! s:cwindow(request, all, copen, mods, event) abort
   endif
   let was_qf = s:is_quickfix()
   let mods = a:mods
-  echom "a:mods = " (mods)
   if mods !~# 'aboveleft\|belowright\|leftabove\|rightbelow\|topleft\|botright'
     let splitdir = g:dispatch_quickfix_split
     let mods = splitdir . ' ' . mods
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -5,7 +5,7 @@ if exists('g:autoloaded_dispatch')
 endif
 
 let g:autoloaded_dispatch = 1
-let g:dispatch_quickfix_split = get(g:, 'dispatch_quickfix_split', 'botright vertical')
+let g:dispatch_quickfix_split = get(g:, 'dispatch_quickfix_split', 'botright')
 
 " Section: Utility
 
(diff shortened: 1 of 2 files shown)


maho:
@tpope any remarks? things to fix?

maho:
@tpope Would you like to take a look at this PR?

tpope:
My general rule is that if I make a plugin command like `:Foo` or `:Xfoo` that is a wrapper of a Vim built-in `:foo`, I don't generally make it configurable in ways that `:foo` is not. So in this case, if `:copen` always open horizontally, then why should `:Copen` be configurable to be vertical? In general I don't want to be responsible for maintaining configuration options that even Vim has rejected.

In this _specific_ case, I have not just philosophical objections but practical ones. With the existing implementation, you can do `:Copen` to get a horizontal split and `:vert Copen` to get a vertical split. With this option in place, you could do `:Copen` to get a vertical split and `:vert Copen` to get the same vertical split. The command is actually less useful with this option.

_Furthermore_, trying it out in practice, it doesn't work consistently. Quickfix windows set `'winfixheight` but not `'winfixwidth'`. That means `:vert copen` creates a narrow window, but the second you do something like close another window, it snaps out wide as `'equalalways'` kicks in. No wonder Vim doesn't have an option to do it automatically: Vim doesn't even work very well when you do it manually.

If you really want this behavior, I would encourage you to use an autocommand that fires after the window opens and move it after the fact. This would also give you an opportunity to set `'winfixwidth'` and make any other necessary tweaks. And it should be possible without a patch to Dispatch.vim.

maho:
ok, I will give autocommand a try. And I was unaware of `:vert Copen`

The reason I did it in such way, is that before `:Copen` I used `:copen` and moved it to right place by `Ctrl-W L` or `Ctrl-W H`. 

`copen` stays in the same position,  while `:Copen` keeps reborning in the same horizontal split position every time `Make` is spawned. Maybe there is easy way to remember position of `:Copen` window?

Horizontal split is nice for vertical displays, but for common horizontal is way less usable.

tpope:
> `copen` stays in the same position, while `:Copen` keeps reborning in the same horizontal split position every time `Make` is spawned. Maybe there is easy way to remember position of `:Copen` window?

I've tentatively pushed a change to avoid unnecessarily closing and re-opening the quickfix window.

PR #334, opened 2022-07-13
Fix interpolated string of ruby

jiz4oh:
Hi @tpope. I encountered an error when I run :Make with interpolated string in ruby. It will expand '#' to alternative file name.

You could reproduce this error by the following:

```console
vim test.rb

:e test2.rb
:Make -e 'a = 1; p "it's #{a}"'
```

Will output
image

And this PR will fix this,

Will output
image


tpope:
Use a `\`. I don't see any reason to add a special case for one piece of syntax for one interpreter.

Issue #339, opened 2023-03-30
How check if the command succeeded?


jclsn:
Is there any way to see if the build (or any other command) succeeded? It is not clear to me from the readme. I would like to use ``:Make`` to build and start debugging afterwards, but only when the build was successful of course.

jclsn:
Maybe I should be more clear. This is what I am trying

```
function! s:BuildAndDebug()
	:Make

	if dispatch#request().status
		autocmd QuickFixCmdPost make ++once call vimspector#Launch()
	endif
endfunction
```

Can you tell me what I am doing wrong?

tpope:
`:Make` is asynchronous. You're checking the exit status before it has exited.

jclsn:
Right, I could have though of that. I got an output now, but there is nothing like "status". There is "completed" though, but that is always 1, if the build fails or not. This is shown as success or failed though.
```
:!cmake --build build -- (job/36615)
{'pid': 36615, 'background': 1, 'action': 'make', 'job': 'process 36615 run', 'program': 'cmake --build build --', 'command': 'cmake --build build --', 'file': '/tmp/vLZNAVp/67', 'id': 63, 'handler': 'job', 'mods': '', 'directory':
 '/home/jan/Workspace/C++/constexpr', 'format': '%*[^"]"%f"%*\D%l: %m,"%f"%*\D%l: %m,%-G%f:%l: (Each undeclared identifier is reported only once,%-G%f:%l: for each function it appears in.),%-GIn file included from %f:%l:%c:,%-GIn f
ile included from %f:%l:%c\,,%-GIn file included from %f:%l:%c,%-GIn file included from %f:%l,%-G%*[ ]from %f:%l:%c,%-G%*[ ]from %f:%l:,%-G%*[ ]from %f:%l\,,%-G%*[ ]from %f:%l,%f:%l:%c:%m,%f(%l):%m,%f:%l:%m,"%f"\, line %l%*\D%c%*[^
 ] %m,%D%*\a[%*\d]: Entering directory %*[`'']%f'',%X%*\a[%*\d]: Leaving directory %*[`'']%f'',%D%*\a: Entering directory %*[`'']%f'',%X%*\a: Leaving directory %*[`'']%f'',%DMaking %*\a in %f,%f|%l| %m', 'expanded': 'cmake --build
build --', 'title': 'make', 'args': ''}
Success: !cmake --build build -- (job/36615)
```

tpope:
[...]

PR #345, opened 2024-02-11
Add usercmd to run after dispatch

Kamilcuk:
This allows users to generically override dispatch with their custom behavior. Some kind of notification is required when the dispatch finishes.

For example, I wanted quickfix window to _always_ open. I have this in my vimrc:

```
	function KcDispatchComplete() abort
		Copen
	        " let the cursor stay where it is.
		normal G
		wincmd p
	endfunction
	autocmd User DispatchComplete call KcDispatchComplete()
```

Amazing plugin, thanks.

### DIFF
--- a/autoload/dispatch.vim
+++ b/autoload/dispatch.vim
@@ -1253,6 +1253,7 @@ function! dispatch#complete(file, ...) abort
     if !a:0
       checktime
     endif
+    doautocmd User DispatchComplete
   endif
   return ''
 endfunction


tpope:
We already call `doautocmd QuickFixCmdPost make` in this case. I've been reticent to add additional hooks for cases that adherence to the Vim standard already covers.

In terms of the problem you're trying to solve, I'm not sure this is the best approach. `:Copen` itself kicks off other autocommands, I feel like you're setting yourself up for jankiness. You can also force the window to open by changing the exit status of the command to nonzero or the `'errorformat'` to match any line of output (e.g., add `%+G%.%#` to the end). I could see an argument for making this easier to do.

tpope:
See also discussion in #339.

PR #346, opened 2024-02-17
Remove redundant variable `winid`

teoljungberg:
It is already set a few further up


teoljungberg:
Here is the original variable: https://github.com/tpope/vim-dispatch/pull/346/files#diff-607b6b923f7b1d85747f0b81fbdb1c1aebe2e6431d11f3c502f059b60db91361R21

PR #351, opened 2024-09-02
Add missing bufnr to s:set_current_compiler()

teoljungberg:
This requirement of bufnr() was introduced in
67a49faf59fa1b41f26e95c46d49c1e4f42e27d5


teoljungberg:
There are probably better ways of fixing this more inline with the plugin by passing through the `a:` arguments - feel free to use this patch :)

PR #363, opened 2025-08-15
Add Kitty support

trekdemo:
The strategy adds support for `:Make`/`:Dispatch` and `:Start`/`:Spawn` task execution using the [Kitty](https://sw.kovidgoyal.net/kitty) terminal emulator's windows and tabs.

`:Make`/`:Dispatch` tasks use kitty's get-text[^1] feature to avoid piping command outputs. This allows interactive debugging sessions and complex TUIs to function. It also allows outputs to display in color during execution and capture the output without ANSI codes.

Environment isolation is implemented using the `--copy-env` and `--env` flags. Environment variables are copied over, except the ones set by (neo)vim itself.

[^1]: https://sw.kovidgoyal.net/kitty/remote-control/#kitten-get-text

https://github.com/user-attachments/assets/4bf1b6b5-0067-4239-9cfa-03c70804fcc0

---

👋 Hey Mr. Pope,

A huge fan here. I've used this plugin for the last decade. I'm grateful for your work and glad that this plugin helped me have fun and earn a living.

I used this plugin with the tmux strategy since the beginning. Lately, I grew apart from tmux and started using the multiplexing capabilities of the Kitty terminal emulator. 

I've been testing this patch for the last month and have fixed all the issues I encountered. It's my first major Vimscript project, so I'm unsure if everything meets the standards.

Please let me know if this strategy is something you would like to add to this esteemed plugin.

### DIFF
--- a/autoload/dispatch/kitty.vim
+++ b/autoload/dispatch/kitty.vim
@@ -56,7 +56,7 @@ function! s:make(request) abort
 
   let cmd = s:kitty_launch_cmd(type, a:request, dispatch#prepare_start(a:request, cmd_with_capturing, 'make'))
   let output = system(cmd)
-  let window_id = matchstr(output, '^\d\+$')
+  let window_id = matchstr(output, '\d\+')
 
   if !empty(window_id)
     let s:waiting[window_id] = a:request
--- a/autoload/dispatch/kitty.vim
+++ b/autoload/dispatch/kitty.vim
@@ -1,0 +1,113 @@
+" dispatch.vim kitty strategy
+"
+" This handler requires you to enable kitty's remote control feature.
+" See https://sw.kovidgoyal.net/kitty/remote-control/
+
+if exists('g:autoloaded_dispatch_kitty')
+  finish
+endif
+let g:autoloaded_dispatch_kitty = 1
+
+let s:waiting = {}
+
+" -----------------------------------------------------------------------------
+" Public handler interface
+
+function! dispatch#kitty#handle(request) abort
+  if empty($KITTY_LISTEN_ON) || !executable('kitten')
+    return 0
+  endif
+
+  if a:request.action ==# 'make'
+    return s:make(a:request)
+  elseif a:request.action ==# 'start'
+    return s:start(a:request)
+  endif
+endfunction
+
+function! dispatch#kitty#activate(pid) abort
+  call system('kitten @ focus-window --match pid:' . a:pid)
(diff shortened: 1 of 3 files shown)


kelvinst:
I have being using it and it works like a charm. Maybe just a good idea to be clearer on the instruction about setting remote control. I mean, I was just setting as `allow_remote_control yes` and was not setting `listen_on` option in `kitty.conf`. After I did that it work perfectly

trekdemo:
> I have being using it and it works like a charm. Maybe just a good idea to be clearer on the instruction about setting remote control. I mean, I was just setting as `allow_remote_control yes` and was not setting `listen_on` option in `kitty.conf`. After I did that it work perfectly

@kelvinst, thank you for the tip. I added an example kitty config with some explanation and a reference to the neovim issue tracker.

autoload/dispatch/kitty.vim
```diff
@@ -0,0 +1,113 @@
...
+    let type = 'window'
+  endif
+
+  let cmd_with_capturing = a:request.expanded .
+                      \ '; echo ' . dispatch#status_var() . ' > ' . a:request.file . '.complete' .
+                      \ '; kitten @ get-text --match=id:$KITTY_WINDOW_ID --ansi=no --extent all > ' . a:request.file
+
+  let cmd = s:kitty_launch_cmd(type, a:request, dispatch#prepare_start(a:request, cmd_with_capturing, 'make'))
+  let output = system(cmd)
+  let window_id = matchstr(output, '^\d\+$')
```
kelvinst: @trekdemo, for some reason, this was not matching anymore after I updated kitty, maybe a trailing `\n`, IDK. Anyway, I think it's safe to just search for the first sequence of digits, no?

```suggestion
  let window_id = matchstr(output, '\d\+')
```

kelvinst: This was causing dispatch to spawn duplicated jobs, because the kitty one was started, but it failed to get the window_id, so this returns exit code 1 and dispatch would then fallback to the other handlers (jobstart, terminal, etc). The symptom for me was that a quickfix was shown besides at the same time of the kitty window, which didn't bother me in the beginning, until I started getting some problem because of the concurrent processes trying to write to the same SQLite database.

trekdemo: @kelvinst thank you for both documenting the bug and providing a fix as well.

I updated the PR/branch with your fix. 👍

